Issue #153, opened 2026-05-05
[Feature] Temporary 2x speed with long hold of input


fivepandasna:
### What problem does this solve?

Some shows and movies have slow or boring parts that some people, especially mobile users, may want to fast forward through. Since they only want to speed through a certain segment of the video, changing the speed many times is not a good solution.

### Proposed solution

A temporary 2x speed (maybe even configurable in speed) gesture, like long hold on mobile, holding pause input (space or left click) on desktop, pause on tv, etc. After the input is stopped, speed returns to normal.

### Platform

All / Not platform-specific

### Additional context

This may conflict with screen lock on mobile, however screen lock implies that the user wouldn’t want to interfere with playback anyways.

PR #187, opened 2026-05-09
Add 2x speed hold for mobile and desktop

fivepandasna:
## Summary
Adds long-press 2x speed hold for mobile and extends the existing TV speed hold to desktop. Includes minor renames to reflect that the speed constant and restore method are now shared across supported platforms.

## Related Issues
- Related to #153 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
- Add `_mobileTemporarySpeedHoldActive` / `_mobileTemporarySpeedRestoreSpeed` state variables for mobile hold tracking
- Add `_canUseMobileTemporarySpeedHold`, `_activateMobileTemporarySpeedHold`, and `_cancelMobileTemporarySpeedHold` to implement long-press 2x speed on mobile
- Wire `onLongPress` / `onLongPressEnd` on the main `GestureDetector` to activate/cancel the mobile speed hold
- Cancel the hide timer during mobile speed hold so the OSD cannot auto-lock while the finger is down; reschedule it on release
- Extend `_canUseTvTemporarySpeedHold` and the key event handler to include `useDesktopUi`, enabling the media key hold for 2× speed on desktop
- Call `_cancelMobileTemporarySpeedHold` in `dispose` and `didChangeAppLifecycleState` to clean up properly
- Rename `_tvTemporarySpeed` → `_temporarySpeed` and `_restoreTvTemporarySpeed` → `_restoreTemporarySpeed` since both are now shared across TV, mobile, and desktop

## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [x] All / Shared code

## Testing

### Test Steps
1. Start playing a video on mobile
2. Long-press the screen and verify playback switches to 2x speed
3. Release and verify speed restores and OSD auto-hides normally
4. On desktop, hold the media pause / play-pause key and verify 2x speed activates; release to verify it restores

N/A

## Checklist

## Notes
There should probably be some sort of indicator for when the video is on 2x speed, maybe near the top of the player. It may also be a good idea to use the space bar/left click on desktop for 2x, but I left it to media keys for now. 

I also had some troubles with unlocking the OSD during testing, however I also experienced these on stable 1.4.0, so I don't think they're related to these changes.

### DIFF
--- a/lib/ui/screens/playback/video_player_screen.dart
+++ b/lib/ui/screens/playback/video_player_screen.dart
@@ -2594,7 +2594,7 @@ class _VideoPlayerScreenState extends State<VideoPlayerScreen>
                 ? () => unawaited(_activateMobileTemporarySpeedHold())
                 : null,
             onLongPressEnd: PlatformDetection.useMobileUi
-                ? (_) => unawaited(_cancelMobileTemporarySpeedHold())
+                ? (_) => _cancelMobileTemporarySpeedHold()
                 : null,
             onDoubleTapDown: PlatformDetection.isTV
                 ? null
--- a/lib/ui/screens/playback/video_player_screen.dart
+++ b/lib/ui/screens/playback/video_player_screen.dart
@@ -59,7 +59,7 @@ class VideoPlayerScreen extends StatefulWidget {
 class _VideoPlayerScreenState extends State<VideoPlayerScreen>
     with WidgetsBindingObserver {
   static final _camelCaseSpaceRe = RegExp(r'(?<=[a-z])(?=[A-Z])');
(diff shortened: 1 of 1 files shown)


fivepandasna:
I thought that builds were fine but I was looking at a workflow run from before I amended a commit, apologies. I fixed the build error and retested on iOS, all is the same. I will test again on Windows and update if there are any issues, but there shouldn't be.

PR #245, opened 2026-05-17
Build - add beta build support for windows (and linux maybe)

broken-droid:
## Summary
add beta build support for windows, and probably linux since it's similar
add an env var in your launch config for MOONFIN_BETA_BUILD to true - this is needed for cmake changes
also pass it in command line with --dart-define - this is for flutter constants in the app for changing paths

added build-windows-beta.ps1 script to handle setting the env and passing --dart-define for building if you want something that runs on command line.

Beta builds will now have separate path locations for their data if you set env and pass the dart-define.

Using build flavors with 'beta' will use these path changes too.

example launch.json for vscode
```
    {
      "name": "Moonfin (Windows - Beta)",
      "request": "launch",
      "type": "dart",
      "program": "lib/main.dart",
      "flutterMode": "debug",
      "deviceId": "windows",
      "env": {
        "MOONFIN_BETA_BUILD": "true"
      },
      "args": [
        "--dart-define=MOONFIN_BETA_BUILD=true"
      ]
    },
```

Windows wasn't able to get dart-define at the cmake level, only in flutter compilation.  So you need to do both steps for windows.
Linux may be able to get them during cmake, but safer to just match windows and use both the env and pass --dart-define to make sure until someone can try.

## Type of Change
- [X] Build/CI change

## Changes Made
- added handling of env var 'MOONFIN_BETA_BUILD' in both windows and linux CMakeLists.txt
- conditionally set different output file properties in Runner.rc
- conditionally change the window title in the application in main.cpp/my_application.cc
- add a static helper in storage_path_service.dart that checks for the 'MOONFIN_BETA_BUILD', from passing --dart-define.  This is used to set different write paths depending on build type.
- added flavor detection, and flavors with 'beta' will use the beta app paths too.
- getOfflineRoot, getDatabaseFile, and getImageCacheDir now use the conditional path string instead of the hardcoded 'Moonfin'
- database_connection_impl.io.dart now uses the static helper that was added to storage_path_service.dart to use a conditional path too

## Platform
- [X] Windows
- [X] Linux
- [X] All / Shared code
Some hardcoded paths were changed, this will affect all platforms if the env var is used in your build configuration.

## Testing
- [X] Not tested (explain why):
only windows was tested, linux might work if someone tries

### Test Steps
1. Add MOONFIN_BETA_BUILD env to launch.json
2. Run/Debug
3. Verify new build exe name, and distinct storage paths for the beta variant

## Checklist

### DIFF
--- a/build-windows-beta.ps1
+++ b/build-windows-beta.ps1
@@ -1,0 +1,29 @@
+param(
+	[string[]]$FlutterArgs = @()
+)
+
+# Set the build flag for this process only
+$env:MOONFIN_BETA_BUILD = "true"
+
+# wipe out the windows build directory if you're running into build problems
+#$buildDir = Join-Path $PSScriptRoot "build\windows\x64"
+#if (Test-Path $buildDir) {
+#	Write-Host "Removing existing Windows build dir: $buildDir"
+#	Remove-Item -Recurse -Force $buildDir
+#}
+
+# Ensure the dart-define is passed to the Flutter tool so `const.fromEnvironment`
+# in Dart will be set at compile time. Also allow additional args to be passed.
+$argsList = @("build", "windows", "--release") + $FlutterArgs
+if (-not ($argsList -contains "--dart-define=MOONFIN_BETA_BUILD=true")) {
+	$argsList += "--dart-define=MOONFIN_BETA_BUILD=true"
+}
+Write-Host "Running: flutter $($argsList -join ' ')"
+& flutter @argsList
+$exit = $LASTEXITCODE
+if ($exit -ne 0) {
+	Write-Error "flutter build windows failed with exit code $exit"
+	exit $exit
+}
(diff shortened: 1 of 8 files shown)


mattsigal:
Hero!

broken-droid:
macos and ios should be able to have a setup like android with flavors.  I can give it a shot if you want, but I can't try it out.

broken-droid:
also, maybe paths should change for windows

getApplicationCacheDirectory - local profile, includes application name
getApplicationDocumentsDirectory - user documents folder
getApplicationSupportDirectory - roaming profile, includes application name

does the db path need to be visible, or are other files going in there too?
Usually cache related things would go in local.  Configs and auth stuff in roaming.

Probably doesn't matter that much and I'm not sure what other platforms need, just a thought

PR #321, opened 2026-06-01
fix: correct Letterboxd rating display (use score/20 instead of value)

mattsigal:
# Pull Request for Letterboxd Rating Fix

## Summary

Fixes Letterboxd ratings displaying as doubled values across all clients.
MDBList's `value` field for Letterboxd is on an ambiguous 0-10 scale
(Letterboxd's native scale is 0-5). The previous `value > 5` halving
guard failed silently for any film scoring **2.5/5 or below** - those
values after doubling are <= 5 and are indistinguishable from a real
low-end score, so they passed through un-halved and displayed incorrectly.

The fix switches to `score / 20`, where `score` is MDBList's always-correct
0-100 normalized field, producing a reliable 0-5 result regardless of
what `value` contains.

## Related Issues

- Fixes known issue of Letterboxd ratings showing doubled values

## Type of Change
- [x] Bug fix

## Changes Made

- **`mdblist_repository.dart`**: Added a `letterboxd`-specific branch in
  the rating value switch that reads `score / 20` instead of `value`.
  Normalizing at the source means all clients will get a correct
  0-5 value automatically.
- **`rating_icon_provider.dart`**: Added a `letterboxd` case to
  `formatRating()` that appends `/5` so the display reads `3.7/5`
  rather than a bare `3.7`.

## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [x] All / Shared code

## Testing

- [x] Manual testing completed - built plugin and tested via web UI as well as on Android TV

### Test Steps
1. Open Moonfin beta on a device with MDBList ratings enabled
2. Navigate to a film with a known Letterboxd score (e.g. *Sorry to Bother You* - 3.7/5 on LB)
3. Confirm the rating badge reads `3.7/5` rather than `7.4` or `7.4/5`
4. Repeat with a poorly-rated film (e.g. a film known to score below 2.5/5) to confirm the lower half of the scale is also correct. Confirmed with The Descent: Part 2 which shows 2.4 (matches Letterboxd.com).

### Screenshots
Current build:
Shield_Screenshot_2026-05-31_22-14-45

Fixed:
Shield_Screenshot_2026-05-31_22-11-49

Low Score (undeserved!):
Shield_Screenshot_2026-05-31_22-29-34

## Checklist

### DIFF
--- a/lib/data/services/rating_icon_provider.dart
+++ b/lib/data/services/rating_icon_provider.dart
@@ -38,6 +38,7 @@ class RatingIconProvider {
       'tomatoes' || 'popcorn' || 'tomatoes_audience' ||
       'tmdb' || 'metacritic' || 'metacriticuser' || 'trakt' || 'anilist' =>
         '${value.toInt()}%',
+      'letterboxd' => '${value.toStringAsFixed(1)}/5',
       _ => value.toStringAsFixed(1),
     };
   }


RadicalMuffinMan:
This was already submitted to the plugin and will be used for all clients, each client doesnt need it

mattsigal:
This handles the formatting of the corrected integer (to display as X/5 instead of just X). This is similar to other ratings that are having a percent sign added to them. If we don't include this, the value is passed to the default _ method, which if I recall, wouldn't handle the value properly. 

I agree that this should ideally happen at the plugin level but for all the ratings - convert everything to strings and pass them as ready to print? But the current PR tried to keep with the logic in place with how the other ratings are handled currently, hence the logic at plugin and formatting at client.

RadicalMuffinMan:
it does double the effort when line 42
`_ => value.toStringAsFixed(1),`
passes all non percent ratings as string literals, its not needed

PR #324, opened 2026-06-01
Moved subtitle selector into it's own file and unit test

TheyCallMeSpy:
# Pull Request

## Summary
Extracts subtitle track selection logic from item_detail_screen.dart into a standalone subtitle_track_logic.dart utility, making it pure and unit-testable. Adds fixture-based unit tests and a fixture loader helper.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

- Extracted isExternalSubtitleStream, isSdhSubtitleStream, computeEffectiveSubtitleIndex, sortedSubtitleStreams, computeSubtitleDialogSelectedIndex, mapSubtitleResultToStreamIndex into lib/util/subtitle_track_logic.dart
- _effectiveSubtitleStreamIndex in item_detail_screen.dart is now a thin wrapper delegating to the pure function
- Added test/util/subtitle_track_logic_test.dart with fixture-based unit tests
- Added test/util/fixture_loader.dart helper
- Added test/fixtures/subtitle_streams.json with real Jellyfin stream data

## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [x] All / Shared code

## Testing

### Test Steps
1. Nothing, just check of subtitle logic is same as before

## Checklist


RadicalMuffinMan:
one small cleanup request: `mapSubtitleResultToStreamIndex()` should guard the lower bound too. Right now a negative result other than 0 would satisfy `result - 1 < displayStreams.length` and then index the list with a negative offset. Something like:
```
final streamPosition = result - 1;
if (streamPosition >= 0 && streamPosition < displayStreams.length) {
  return displayStreams[streamPosition]['Index'] as int?;
}
```
Also, the `cs does not match ces` test doesn't really prove that behavior because the filtered list only contains the Czech stream, so both fallback and a real match return the same index. I’d either rename it as a fallback test or add another stream before the Czech stream so the expected values differ.

RadicalMuffinMan:
This PR was superseded by #639 and is now closed

PR #345, opened 2026-06-03
[codex] Fix desktop player refocus handling

stoopboop:
## Summary

- Keep the media_kit video surface from pausing when desktop apps enter background/unfocused states.
- Add a desktop-only fallback hardware keyboard handler while the video player screen is mounted so playback shortcuts still work if the player focus node is lost after window refocus.

## Why

On macOS, unfocusing the app can leave audio continuing while the rendered video surface stops updating. The desktop Video widget was configured to pause on background entry even though playback itself continues through libmpv. After refocusing, keyboard shortcuts can also stop working if Flutter focus is not restored to the player overlay.

## Validation

- git diff --check passed.
- Attempted targeted dart analyze, but this local checkout has unresolved package dependencies, including Flutter package imports, so analysis could not produce useful patch-specific results.


RadicalMuffinMan:
desktop player refocus was fixed in v2.0 and the PR template was not used

PR #346, opened 2026-06-03
[codex] Fix TV add-server keyboard dismissal

stoopboop:
## Summary

- Add a non-submitting keyboard hide path to CustomTVTextField.
- Hide the TV keyboard and move focus to the Connect button before/after add-server connection attempts.
- Explicitly hide the system IME when deactivating a system keyboard session.

## Why

On Android/Google/TCL TV devices, connection errors while entering a server address can leave the on-screen keyboard visible or focused while the dialog shows an error. That traps remote navigation in the keyboard layer and makes it difficult to hide the keyboard or reach dialog actions.

## Validation

- git diff --check passed.
- Attempted targeted dart analyze, but this local checkout has unresolved package dependencies, including Flutter package imports, so analysis could not produce useful patch-specific results.

### DIFF
--- a/lib/ui/screens/auth/server_select_screen.dart
+++ b/lib/ui/screens/auth/server_select_screen.dart
@@ -918,6 +918,10 @@ class _AddServerDialogState extends State<_AddServerDialog> {
   Future<void> _submit() async {
     final address = widget.controller.text.trim();
     if (address.isEmpty) return;
+    if (PlatformDetection.isTV) {
+      _tvFieldKey.currentState?.hideKeyboard();
+      _connectFocus.requestFocus();
+    }
 
     setState(() {
       _isConnecting = true;
@@ -944,7 +948,9 @@ class _AddServerDialogState extends State<_AddServerDialog> {
       _errorMessage = message;
     });
     WidgetsBinding.instance.addPostFrameCallback((_) {
-      if (mounted) _connectFocus.requestFocus();
+      if (!mounted) return;
+      _tvFieldKey.currentState?.hideKeyboard();
+      _connectFocus.requestFocus();
     });
   }
 
--- a/packages/custom_tv_text_field_fork/lib/src/custom_tv_text_field.dart
+++ b/packages/custom_tv_text_field_fork/lib/src/custom_tv_text_field.dart
@@ -292,8 +292,12 @@ class CustomTVTextFieldState extends State<CustomTVTextField>
   }
 
   void closeKeyboard() {
+    hideKeyboard(submit: true);
+  }
+
+  void hideKeyboard({bool submit = false}) {
     if (_keyboardController.isVisible) {
-      _keyboardController.hide(true);
+      _keyboardController.hide(submit);
       return;
     }
     _deactivateSystemIme();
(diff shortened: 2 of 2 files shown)


RadicalMuffinMan:
Hi! Thanks for the submission. The TV keyboard trap on add-server errors is a real annoyance, and the idea of hiding the keyboard and moving focus to Connect on both the connect and error paths is the right approach.

 A few things before this can go in:

The main concern is that CustomTVTextField is not a TV only widget. Despite the name, it's the app's shared adaptive text field, used in login, search, Seerr config, syncplay, add-to-collection, Emby connect, and more. On mobile and desktop the native keyboard is exactly this widget's system IME path: preferSystemImeKeyboard defaults to true there (the toggle to disable it is TV monly), so those platforms run with _useSystemImeSession active by default. The change to _deactivateSystemIme() adds SystemChannels.textInput.invokeMethod('TextInput.hide'), and that method is called from _onSystemInputFocusChanged on routine focus loss, not just explicit close. So this now force hides the system keyboard every time the field loses focus, on mobile and desktop too. The likely side effect is that when a user moves from one field to the next, say username to password on login, or tapping the search box, the first field's blur fires TextInput.hide and then the next field re-shows it, so you get a hide/show flicker or an unwanted dismissal where today the keyboard just stays up. That is a cross-platform regression coming out of a TV specific fix.

Could you scope the forced IME-hide so it only runs on the explicit close paths (hideKeyboard and closeKeyboard), or gate it to TV? That keeps your fix while leaving mobile keyboard behavior unchanged. One minor note, the _systemInputFocusNode.unfocus() is redundant on the blur path, since the node has already lost focus there.

There's also a small inconsistency. _submit() is gated on PlatformDetection.isTV, but the error-handler callback calls hideKeyboard() unconditionally, so it runs on mobile too since the dialog builds a CustomTVTextField there. Probably harmless, but worth making the two paths consistent.

On testing, since this touches a widget used app-wide, it really needs a device pass on both sides: the TV add server error flow it fixes, and mobile field-to-field keyboard behavior in login and search to confirm there's no flicker. The PR notes that dart analyze couldn't run locally and there's no build or test check on the PR, so it's currently unverified. This branch is also a few weeks old and a fair amount of keyboard and focus work has landed on main since, so it's worth confirming the trap still reproduces before finalizing.

Last thing, the PR didn't use the repo's pull request template. Could you fill it in, especially Type of Change, Platform, which should have several boxes checked rather than just TV, and the Testing and Checklist sections? Given the cross-platform reach here, the Platform and Testing boxes matter for this one in particular. Thank you!

stoopboop:
Thanks for taking a look. That makes sense.

My main intent here was to report the TV add-server keyboard trap rather than push a specific implementation. I agree this PR probably reaches too far into shared keyboard behavior, and I do not have the device/testing coverage to validate a proper cross-platform fix.

I am going to close this PR so it does not look like an active code contribution. The useful signal is the issue itself: on TV, after an add-server connection error, the on-screen keyboard can remain active/focused in a way that traps remote navigation and makes it hard to reach the dialog actions.

If you would prefer this tracked as a separate issue instead of a closed draft PR, I can open one with the behavior and repro context.

PR #347, opened 2026-06-03
[codex] Smooth toolbar label expansion animation

stoopboop:
## Summary

- Keep toolbar button labels mounted while animating expansion/collapse.
- Animate label width and opacity instead of inserting/removing the label subtree during hover/focus changes.
- Slightly lengthen the expansion duration for a less abrupt transition.

## Why

The top toolbar currently changes button width while also adding/removing the text label from the widget tree. On hover or focus changes this can produce visible flashing and uneven size transitions. Keeping the label subtree stable makes the animation closer to the cleaner tvOS-style behavior without replacing the whole navbar implementation.

## Validation

- git diff --check passed.
- Attempted targeted dart analyze earlier in this checkout, but package dependencies are unresolved locally, including Flutter package imports, so analysis could not produce useful patch-specific results.

### DIFF
--- a/lib/ui/widgets/expandable_icon_button.dart
+++ b/lib/ui/widgets/expandable_icon_button.dart
@@ -10,7 +10,7 @@ import '../../util/focus/dpad_keys.dart';
 import '../../util/platform_detection.dart';
 import 'focus/focus_theme.dart';
 
-const _kExpandDuration = Duration(milliseconds: 150);
+const _kExpandDuration = Duration(milliseconds: 220);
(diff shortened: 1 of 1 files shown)


RadicalMuffinMan:
Hi, and thanks for putting this together. 

I built it and compared it against main, side by side on desktop and on Android TV. The honest result is that the difference is very hard to perceive. On TV the toolbar labels only expand on focus, and the old pop versus the new slide is subtle enough that it's barely distinguishable in normal use, and it's pretty subtle on desktop hover too. I don't think I'll be moving forward with this addition, thank you for your time.

PR #351, opened 2026-06-03
feat(home page): Implement horizontal infinite scrolling for home rows

mattsigal:
# Pull Request: The Lazy Loader

## Summary
Add seamless horizontal infinite scrolling (lazy loading) for all paginatable built-in home screen rows (Playlists, Favorites, Collections, Genres, and Latest Media) under both single-server and multi-server library modes. Bonus: ensure they are properly linked to their Media Details view.

## Related Issues
- Related to Randall's request

## Type of Change
- [x] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
- **Initial Limits**: Committed to initial load limit to `_defaultLimit` (15) for core and custom collection, genre, and favorite row loading methods.
- **Scroll Pagination Trigger**: Updated `LockedFocusRow` in `home_screen.dart` to automatically trigger `loadMoreForRow` when the focused item index is at the end of the row.
- **Unified Offset Tracking**: Integrated offset pagination for `loadMore` queries in `RowDataSource` and `MultiServerRepository` (aggregating query offsets per active server using the current count of items from that server in the combined row).
- **Offset Alignment (Duplicates Fix)**: Aligned page-based offsets to multiples of the default page size (`_defaultLimit`) for `Playlists` and `Genres` queries to prevent item overlaps caused by local client-side filtering.
- **Stable ID Parsing Fix**: Switched `_parseStableId` parsing to a right-to-left method (`lastIndexOf(':')`) to support server URLs containing colons (e.g. `http://...`) without corrupting genre/collection IDs.
- **Genres/Aggregated Rows `totalCount` Fix**: Fixed `totalCount` properties in `RowDataSource.loadGenres` and aggregated row builders to correctly enable scroll pagination.

## Platform
- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [x] All / Shared code

## Testing

### Test Steps
1. Perform clean build and deploy the `androidTv-beta` APK to Nvidia Shield TV.
2. Scroll through home screen rows (Genres, Latest Media, Playlists, Favorites).
3. Scroll horizontally near the end of any custom or core row to verify that additional items load and append seamlessly without duplicates.

### Screenshot
Showing the end of the Genres row (27 items):
Shield_Screenshot_2026-06-03_01-21-31

## Checklist

### DIFF
--- a/lib/data/repositories/multi_server_repository.dart
+++ b/lib/data/repositories/multi_server_repository.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 
+import 'package:get_it/get_it.dart';
 import 'package:logger/logger.dart';
 import 'package:server_core/server_core.dart';
 
@@ -7,6 +8,8 @@ import '../../auth/models/server.dart';
 import '../../auth/repositories/session_repository.dart';
 import '../../auth/store/authentication_store.dart';
 import '../../auth/store/credential_store.dart';
+import '../../preference/preference_constants.dart';
+import '../../preference/user_preferences.dart';
 import '../models/aggregated_item.dart';
 import '../models/aggregated_library.dart';
 import '../models/home_row.dart';
@@ -50,6 +53,7 @@ class MultiServerRepository {
   static const _imageTypes = 'Primary,Backdrop,Thumb';
   static const _imageTypeLimit = 1;
   static const _defaultLimit = 15;
+  static const _maxItems = 100;
   static const _defaultSortBy = 'SortName';
   static const _defaultSortOrder = 'Ascending';
   static const _genreArtworkConcurrency = 6;
@@ -278,11 +282,15 @@ class MultiServerRepository {
     final all = results.expand((e) => e).toList()
       ..sort((a, b) => a.name.compareTo(b.name));
 
+    final takenItems = all.take(limit).toList();
+    final totalCount = takenItems.length < limit ? takenItems.length : _maxItems;
+
     return HomeRow(
       id: 'playlists',
       title: _l10n.playlists,
-      items: all.take(limit).toList(),
+      items: takenItems,
       rowType: HomeRowType.playlists,
+      totalCount: totalCount,
     );
   }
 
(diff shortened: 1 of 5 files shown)


skank01:
Would love to see infinite row scrolling!
Also using the modern style... is it possible the have the big cover (one that is selected) at the start of the row, and keep it there, not moving to the end of the row?

mattsigal:
> Would love to see infinite row scrolling! Also using the modern style... is it possible the have the big cover (one that is selected) at the start of the row, and keep it there, not moving to the end of the row?

It only goes to the right-side if you make it to the final card of the (infinite scrolled) content. Otherwise the focused card is the left-most.

skank01:
> > Would love to see infinite row scrolling! Also using the modern style... is it possible the have the big cover (one that is selected) at the start of the row, and keep it there, not moving to the end of the row?
> 
> It only goes to the right-side if you make it to the final card of the (infinite scrolled) content. Otherwise the focused card is the left-most.

So if the row is infinite like you made here... it will stay on the left side? That would make sense!

mattsigal:
> > > Would love to see infinite row scrolling! Also using the modern style... is it possible the have the big cover (one that is selected) at the start of the row, and keep it there, not moving to the end of the row?
> > 
> > 
> > It only goes to the right-side if you make it to the final card of the (infinite scrolled) content. Otherwise the focused card is the left-most.
> 
> So if the row is infinite like you made here... it will stay on the left side? That would make sense!

Again, it is infinite in the sense that if the row's source has 500 items, it'll keep pulling them until you have all 500 visible. There is no upper limit. However, it is not infinite in that it will "loop" and continue scrolling to the right once you hit the full set (which is what is visible in my screenshot - I came to the last genre and that's the end of the row).

skank01:
> > > > Would love to see infinite row scrolling! Also using the modern style... is it possible the have the big cover (one that is selected) at the start of the row, and keep it there, not moving to the end of the row?
> > > 
> > > 
> > > It only goes to the right-side if you make it to the final card of the (infinite scrolled) content. Otherwise the focused card is the left-most.
> > 
> > 
> > So if the row is infinite like you made here... it will stay on the left side? That would make sense!
> 
> Again, it is infinite in the sense that if the row's source has 500 items, it'll keep pulling them until you have all 500 visible. There is no upper limit. However, it is not infinite in that it will "loop" and continue scrolling to the right once you hit the full set (which is what is visible in my screenshot - I came to the last genre and that's the end of the row).

ok i thought that if its entered the last one of the row's source, say 500 or 100, it would continue to the first one again.. 
That way you could keep the focus with the big one on the left.
So not scrolling to the right.. the row just populates infinite (say 500)  and after that one, it starts with number one again.

You know what i mean...
Cause if you have 500 like you've set it up now.. what happens , cause you are focussed to the most right part, then you have to go left again?

lib/data/repositories/multi_server_repository.dart
```diff
@@ -278,11 +282,15 @@ class MultiServerRepository {
     final all = results.expand((e) => e).toList()
       ..sort((a, b) => a.name.compareTo(b.name));
 
+    final takenItems = all.take(limit).toList();
+    final totalCount = takenItems.length < limit ? takenItems.length : _maxItems;
```
RadicalMuffinMan: This line decides whether a row "has more items to load." Right now, whenever the first batch comes back full, we set the total to 100 basically a guess. The problem: if the server only actually has 30 items, the app still believes there are 100, so it keeps trying to load more that aren't there.

The good news is the server already tells us the real number! It comes back in the response as TotalRecordCount. If we use that instead of 100, the app will know exactly when to stop. This same line shows up a few more times in this file (for genres, generic rows, and latest) so they'd all get the same fix.

lib/data/repositories/multi_server_repository.dart
```diff
@@ -360,14 +368,173 @@ class MultiServerRepository {
...
       id: 'genres',
       title: _l10n.genres,
-      items: all.take(limit).toList(),
+      items: takenItems,
       rowType: HomeRowType.genres,
+      totalCount: totalCount,
     );
   }
 
+  FavoriteTypeFilter _favoriteFilterForId(String id) {
```
RadicalMuffinMan: there's a nearly identical version of this little lookup in another file (row_data_source.dart, called _favoriteItemTypesForId). When the same logic lives in two places, they tend to slowly drift apart and cause bugs and these two already behave a little differently in their fallback case. Could we move this to one shared spot so there's a single source of truth?

lib/data/repositories/multi_server_repository.dart
```diff
@@ -360,14 +368,173 @@ class MultiServerRepository {
...
+            if (row.id.startsWith('latest_')) {
+              final parts = row.id.split('_');
+              if (parts.length >= 3) {
+                final rowServerId = parts[1];
+                final parentId = parts[2];
+                if (serverId != rowServerId) return const <AggregatedItem>[];
+
+                final response = await session.client.itemsApi.getLatestItems(
+                  parentId: parentId,
+                  limit: row.items.length + _defaultLimit,
```
RadicalMuffinMan: Small thing about how "Latest" rows load: unlike the other rows, this endpoint can't ask for "the next 15." So each time it loads more, it re-downloads everything from the start with a bigger limit. That's okay-ish, but right now there's nothing telling it when to stop, so once it's loaded everything it'll keep re-downloading the same list. add a check: "if we didn't get any new items, we're done"

lib/data/repositories/multi_server_repository.dart
```diff
@@ -360,14 +368,173 @@ class MultiServerRepository {
...
+            }
+            return const <AggregatedItem>[];
+          default:
+            return const <AggregatedItem>[];
+        }
+      }, label: 'loadMore ${row.rowType} from ${session.server.name}')),
+    );
+
+    final newItems = results.expand((e) => e).toList();
+    if (newItems.isEmpty) return row.items;
```
RadicalMuffinMan: This sends back just the list of items, but it doesn't tell the caller "by the way, that was the last page." Because of that, the rest of the app can't know to stop asking. If this returned a little extra info (like the real total, or a simple "any more left?" yes/no), the calling code could do the right thing. Tied to the first comment above on line 383.

lib/data/services/row_data_source.dart
```diff
@@ -598,31 +607,228 @@ class RowDataSource {
     );
   }
 
+  List<String>? _favoriteItemTypesForId(String id) {
```
RadicalMuffinMan: This is the twin of _favoriteFilterForId over in multi_server_repository.dart, same idea, written twice. Worth combining them into one shared helper so they can't get out of sync (they're already slightly different).

lib/data/services/row_data_source.dart
```diff
@@ -598,31 +607,228 @@ class RowDataSource {
...
+          final parentId = row.id.substring('latest_'.length);
+          final response = await _getLatestItemsWithFallback(
+            parentId: parentId,
+            limit: row.items.length + _defaultLimit,
+          );
+          final items = normalizeLatestMediaItems(
+            _parseItems(response, serverId),
+            limit: row.items.length + _defaultLimit,
+          );
+          return items;
```
RadicalMuffinMan: Same note as the "Latest" rows in the other file: it re-downloads the whole list each time and never decides it's finished. A quick "did we actually get anything new? if not, stop" check would do it.

lib/data/services/row_data_source.dart
```diff
@@ -598,31 +607,228 @@ class RowDataSource {
...
           startIndex: row.items.length,
           limit: _defaultLimit,
         );
+      case HomeRowType.genres:
+        final sortBy = prefs?.get(UserPreferences.genresRowSortBy).apiValue ?? _defaultSortBy;
+        final includeItemTypes = prefs?.get(UserPreferences.genresRowItemFilter).includeItemTypes;
+        final parsed = _parseStableId(row.id);
+        if (row.id == 'genres') {
+          final browseItemTypes = normalizeBrowsableGenreItemTypes(includeItemTypes);
+          final pageCount = (row.items.length / _defaultLimit).ceil();
```
RadicalMuffinMan: possible edge case to be aware of: if a whole batch happens to get filtered out, the counter doesn't move forward and it can get stuck asking for the same batch over and over.

RadicalMuffinMan (changes requested):

lib/data/repositories/multi_server_repository.dart
mattsigal: Can do

mattsigal:
All comments addressed!

1. Moved all row ID to type filter mapping logic into FavoriteTypeFilter.fromRowId(id) inside preference_constants.dart and removed local lookups from row_data_source.dart and multi_server_repository.dart.
2. Added a _rowOffsets tracking map in HomeViewModel (and per-server tracking in MultiServerRepository.loadMore) to advance startIndex values independently of list lengths. If a page gets completely filtered out by client-side filters, the query counter still moves forward, avoiding the infinite request loop.
3. Swapped out hardcoded total count assumptions (100) for Dart 3 records (List<AggregatedItem>, int) mapping the exact TotalRecordCount returned by the server.
4. Implemented a check on "Latest" rows: if the returned list size doesn't increase, paging is stopped immediately.

Issue #209, opened 2026-05-11
[Playback] Cinema mode playback only works for user 1


skank01:
### Moonfin version

1.9.1

### Platform

Android

### OS version

Android 14

### Device model

Happens on multiple devices, google chromecast, shield tv

### Input method

Touch

### Server type and version

Latest

### Content type

Movie

### Media details (as much as you know, check the info button at the bottom right of the player)

 I just tested on my shield... same issue with the cinema mode. like on chromecast.. if i enable it (then its enabled for all 4 users/profiles on the device. It plays perfect for user 1 , but for user 2,3,4 movies arent played... (it doesnt even start with the preroll).... As soon as i disable cinema mode, all movies play for 4 users

Been present since version 1.8

### Playback path

Direct Play

### Symptoms (check all that apply)

- [x] Black screen / endless spinner
- [ ] Audio only (no video)
- [ ] Stuttering / frame drops
- [ ] A/V sync issues
- [ ] Seeking/scrubbing broken
- [ ] Subtitles missing or incorrect
- [ ] Unexpected forced transcode
- [ ] App crashes / exits

### Steps to reproduce

Log in user 1, enable cinema mode
Testplay movie -> all ok
Log out
Log in user 2
check that cinema mode is enabled too (it should be cause you cant disable for 1 user, so enabled for all users, of disabled for all users)
Start movie, it spins and stops

### Frequency

Always

skank01:
Still happening on 1.9.2
Even worse.
User 1 plays ok
User 2 do3s nothing when hitting play movie button when cinrma mode enabled

RadicalMuffinMan:
This has been fixed for the upcoming release

skank01:
Still not fixed, not in 2.0.0 and not in 2.0.1
Hitting play with other user than user 1 , it does nothing... when cinema mode enabled

PR #371, opened 2026-06-05
Cinema Mode Safety Check & Transition Fixes

mattsigal:
# Pull Request for Cinema Mode Safety Check & Transition Fixes

## Summary
Implements a safety check and transition fix for Cinema Mode pre-rolls. This prevents playback from hanging on a black screen when pre-rolls fail to load (which can happen on restricted secondary accounts) or during transitions from a pre-roll to the main movie.

## Related Issues
- Closes issue [209](https://github.com/Moonfin-Client/Moonfin-Core/issues/209)

## Type of Change
- [x] Bug fix
- [x] UI/UX update
- [x] Refactor

## Changes Made
- **Error Propagation & Watchdog Auto-Skip:**
  - Added an `errorStream` to `MediaKitPlayerBackend` to expose underlying media kit errors.
  - Implemented error handling and queue fallback in `PlaybackManager`. If a pre-roll fails to play or load, the player cleanly stops, suppresses the playback error prompt, advances the queue, and automatically starts the main movie.
- **Stale State Management during Transitions:**
  - Implemented an `_isStale` boolean flag in `MediaKitPlayerBackend`. During stop/play transitions, `_isStale` remains true until the media player playlist active index matches the loaded target URL.
  - When `_isStale` is true, position/duration/buffering streams and getters return zero/false, preventing incorrect UI state updates or transition hangs.
  - Modified `waitForTracksReady` and `waitForEmbeddedSubtitleCount` to yield immediately if the player state is stale.

## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [x] Windows
- [ ] Linux
- [x] All / Shared code

## Testing
- [x] Manual testing completed - verified on Windows 11 Desktop across primary and secondary accounts

### Test Steps
1. Log in to a secondary account that does not have permissions or access to pre-rolls, and play a movie with Cinema Mode enabled.
2. Verify that the player gracefully skips the pre-roll immediately without showing a black screen or throwing an error, playing the movie.
3. Log in to a primary account, play a movie with Cinema Mode enabled, and verify that the pre-roll plays and transitions to the main feature automatically without hanging.

## Checklist

### DIFF
--- a/packages/playback_core/lib/src/playback_manager.dart
+++ b/packages/playback_core/lib/src/playback_manager.dart
@@ -940,16 +940,7 @@ class PlaybackManager {
       } catch (_) {}
 
       if (_isPreroll(item)) {
-        final hadNext = queueService.next();
-        if (hadNext) {
-          await _playCurrentItem(
-            startPosition: startPosition,
-            enableDirectPlay: enableDirectPlay,
-            enableDirectStream: enableDirectStream,
-            enableTranscoding: enableTranscoding,
-            allowStartupRecovery: allowStartupRecovery,
-          );
-        }
+        await next();
         return;
       }
 
--- a/packages/playback_core/lib/src/playback_manager.dart
+++ b/packages/playback_core/lib/src/playback_manager.dart
@@ -484,11 +484,20 @@ class PlaybackManager {
   }
 
   Future<void> _handleBackendErrorEvent(Map<String, dynamic> event) async {
+    final queueItem = queueService.currentItem;
+    if (_isPreroll(queueItem)) {
+      _suppressNextGenericBackendError = true;
+      try {
+        await _backend!.stop();
+      } catch (_) {}
+      await next();
+      return;
+    }
+
     final eventType = event['event']?.toString();
     final kind = event['kind']?.toString();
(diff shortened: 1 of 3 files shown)


packages/playback_core/lib/src/playback_manager.dart
```diff
@@ -915,6 +939,20 @@ class PlaybackManager {
         await _backend!.stop();
       } catch (_) {}
 
+      if (_isPreroll(item)) {
```
RadicalMuffinMan, line 942: replace this with 
```
if (_isPreroll(item)) {
    await next();
    return;
}
```

The forwarded parameters are wrong for the main movie. Those values (startPosition, enableDirectPlay, etc.) describe how the pre-roll was being played. Passing them into the main movie's startup means the pre-roll's failed playback context bleeds into the feature. next() starts the next item with clean defaults, which is what you want after skipping a broken pre-roll.

RadicalMuffinMan (changes requested):

packages/playback_core/lib/src/playback_manager.dart
mattsigal: Good catch, sorry about that. Fixed!

PR #373, opened 2026-06-05
Interface Language Override

mattsigal:
# Interface Language Override

## Summary
Implements a user-scoped Interface Language override setting under a new "Account Preferences" settings category. Also fixes a MSVC build compatibility issue with legacy coroutine warnings in third-party plugins.

## Related Issues
- Resolves issue requesting user interface language override preference.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Performance improvement
- [x] UI/UX update
- [ ] Documentation update
- [x] Build/CI change
- [ ] Other (describe):

## Changes Made
  - Registered `languageOverride` explicitly typed as `Preference<String>` in `UserPreferences`.
  - Defined profile-scoped getters/setters for `languageOverride` resolving keys per user (e.g. `pref_language_override_${serverId}_${userId}`).
  - Fixed `PreferenceBinding` constructor/setter to query and write values through the `UserPreferences` wrapper instead of bypassing it to the raw `PreferenceStore`, ensuring scoping works for all preference tiles.
  - Imported `locale_names.dart` with a database of native display names for all 72 supported locales.
  - Resolved locale at startup/rebuild in `app.dart` to inject selected language tag into `MaterialApp.router`.
  - Triggered reactive rebuilds `setState` in root `_AppState` on preference changes.
  - Added new "Account Preferences" section in `settings_side_panel.dart`.
  - Positioned "Interface Language" dropdown first, sorting options alphabetically by native names.
  - Placed "Sort Servers By" immediately under it.
  - Added `-D_SILENCE_EXPERIMENTAL_COROUTINE_DEPRECATION_WARNINGS` in `windows/CMakeLists.txt` to resolve legacy compilation errors in third-party plugins under modern MSVC.

## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [x] All / Shared code

## Testing
- [x] Tested on physical device
- [x] Manual testing completed

### Test Steps
1. Navigate to Settings -> Account & Security.
2. Select a language from "Interface Language" under the new "Account Preferences" section.
3. Verify the app language updates reactively without restarting.
4. Log out and switch to another user profile. Verify that the language preference resets to the system default, demonstrating that account settings are scoped properly.

## Screenshots
System Default (default) and the List of Available Options
2026-06-04_23-38-35_moonfin
2026-06-04_23-38-53_moonfin

Other Selections
2026-06-04_23-39-16_moonfin
[...]

PR #378, opened 2026-06-05
Fix PIN Security and Login Preferences

mattsigal:
# Fix PIN Security and Login Preferences

## Summary
Implements PIN security checks when switching profiles in-app, bypasses immediate PIN verification locks on startup (routing instead to the user selection screen), and renames the auto-login option "Specific User" to "Current User" with dynamic backward compatibility.

## Related Issues
- Issue reported on Discord

## Type of Change
- [x] Bug fix
- [x] UI/UX update

## Changes Made
- **PIN Verification on Profile Switch**: In `user_menu_dialog.dart`, integrated a PIN check utilizing `PinEntryDialog` prior to switching the active session. If "Forgot PIN" is selected, the dialog closes and navigates to that user's login screen.
- **PIN Auto-Login Bypass on Startup**: In `startup_screen.dart`, if the configured auto-login target user has a PIN enabled, auto-login is bypassed and the app routes directly to the server's user selection screen (`ServerScreen`) instead of prompting for the PIN immediately, preventing other users from being locked out.
- **Renamed Auto-Login UI Text**: Renamed the enum value `specificUser` to `currentUser` in `UserSelectBehavior` and updated the UI label to `"Current User"` in English locale files.
- **Backward Compatibility Mapping**: Added dynamic mapping in `PreferenceStore._getEnumDynamic` to automatically translate saved `'specificUser'` strings in `shared_preferences` to `'currentUser'`.

## Platform
- [x] All / Shared code

## Testing
- [x] Tested on physical device (Windows client built locally)

### Test Steps
1. Enable PIN code on a user profile.
2. Select "Current User" under Auto Login in settings.
3. Switch profiles using the profile selection dialog in-app; verify that the PIN entry dialog correctly prompts for authentication and successfully logs in when correct, or aborts/goes to login when canceled/forgotten.
4. Restart the app; verify that instead of prompting immediately for the last user's PIN, it loads the server's user selection screen.
5. Tap the PIN-enabled user and verify that it prompts for the PIN.
6. Verify that a test merge with PR #373 (Language Override) merges cleanly and builds successfully.

## Screenshots
From Switch User screen
2026-06-05_11-58-17_moonfin
From Login screen
2026-06-05_11-58-43_moonfin

## Checklist

### DIFF
--- a/packages/preference/lib/src/store/preference_store.dart
+++ b/packages/preference/lib/src/store/preference_store.dart
@@ -133,8 +133,9 @@ class PreferenceStore {
     try {
       var stored = _requirePrefs.getString(preference.key);
       if (stored == null || stored.isEmpty) return preference.defaultValue;
-      if (stored == 'specificUser') {
+      if (preference.key == 'pref_auto_login_behavior' && stored == 'specificUser') {
         stored = 'currentUser';
+        _requirePrefs.setString(preference.key, 'currentUser');
       }
       for (final v in preference.values) {
         if ((v as Enum).name == stored) return v;
--- a/packages/preference/lib/src/store/preference_store.dart
+++ b/packages/preference/lib/src/store/preference_store.dart
@@ -131,8 +131,11 @@ class PreferenceStore {
 
   dynamic _getEnumDynamic(EnumPreference<dynamic> preference) {
     try {
-      final stored = _requirePrefs.getString(preference.key);
+      var stored = _requirePrefs.getString(preference.key);
       if (stored == null || stored.isEmpty) return preference.defaultValue;
+      if (stored == 'specificUser') {
+        stored = 'currentUser';
+      }
       for (final v in preference.values) {
         if ((v as Enum).name == stored) return v;
       }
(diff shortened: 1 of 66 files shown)


packages/preference/lib/src/store/preference_store.dart
```diff
@@ -131,8 +131,11 @@ class PreferenceStore {
 
   dynamic _getEnumDynamic(EnumPreference<dynamic> preference) {
     try {
-      final stored = _requirePrefs.getString(preference.key);
+      var stored = _requirePrefs.getString(preference.key);
       if (stored == null || stored.isEmpty) return preference.defaultValue;
+      if (stored == 'specificUser') {
```
RadicalMuffinMan: If any other current or future enum ever has a value named specificUser, it gets silently remapped to currentUser regardless of context. It should be gated on the preference key:

```
if (preference.key == 'pref_user_select_behavior' && stored == 'specificUser') {
  stored = 'currentUser';
}
```

packages/preference/lib/src/store/preference_store.dart
```diff
@@ -131,8 +131,11 @@ class PreferenceStore {
...
+        stored = 'currentUser';
```
RadicalMuffinMan, line 137: The remap translates specificUser to currentUser in memory on every read, but the value in shared_preferences stays 'specificUser' permanently. That means this workaround runs on every single preference read for the rest of the app's lifetime, even after no users have the old value stored. A proper migration would write 'currentUser' back to storage the first time it sees 'specificUser', so the mapping code can eventually be removed.

RadicalMuffinMan (changes requested):

PR #390, opened 2026-06-06
Fixing Home Page Issues

### LINKED ISSUE #388: [Playback] Resuming episodes from "Continue Watching" on home page does not autoplay next EP and does not allow next ep navigation in player
### Moonfin version

2.0.1

### Platform

Android

### OS version

Android 16, EMUI15, Windows 11

### Device model

Samsung Tab S9, Huawei Mate 80 Pro, Dell desktop

### Input method

Touch

### Server type and version

Jellyfin 10.11.10

### Content type

Episode

### Media details (as much as you know, check the info button at the bottom right of the player)

When clicking on an episode from continue watching on the home page, it brings me to the details page where I click resume, in the player it doesn't have a next button to go to the next episode nor does it automatically go to the next ep at the end. 
But if I navigate to the series page to resume from there the next button shows up in the player, same episode same resume point. Autoplay is on, next up popup is off.
Same behavior on phone, tablet and windows app. All on moonfin 2.0.1.

Details page that shows "next up" brings the player up with the next button, clicked from series page. 
Details page with chapters is from home page "continue watching"

I don't have HSS or Kefintweaks plugins.

Image
Image
Image
Image

### Playback path

Direct Play

### Symptoms (check all that apply)

- [ ] Black screen / endless spinner
- [ ] Audio only (no video)
- [ ] Stuttering / frame drops
- [ ] A/V sync issues
- [ ] Seeking/scrubbing broken
- [ ] Subtitles missing or incorrect
- [ ] Unexpected forced transcode
- [ ] App crashes / exits

### Steps to reproduce

1. Play episode from homepage "Continue Watching"
2. Press play on details page
3. see no option to click next
4. Go to same episode from series-> season. Or even from homepage "continue watching" -> details page-> "Go to Series" -> Play
5. player gives "Next" button

### Frequency

Always

### Logs / errors (optional)

_No response_

mattsigal:
# Resolving Media Detail Page Discrepancies

## Summary
This PR resolves metadata loading and play/resume state displays on show details pages, specifically for items navigated from or played via "Recent TV Shows", "Recent Movies", and "Continue Watching" rows. It addresses issue #388 by ensuring the details page correctly displays "Play" vs "Resume from SX:EX" and has the user-specific state needed to build the Next Up/resume episode queues.

## Related Issues

- Closes #388 
- Fixes #
- Related to #

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [X] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

- Updated `getItem(itemId)` to use the user-specific `/Users/$userId/Items/$itemId` endpoint instead of the public `/Items/$itemId` to load correct user-specific play and resume data.
- Added `RecursiveItemCount`, `ChildCount`, and `SpecialFeatureCount` to `kItemFields` to determine the accurate played and episode counts for show/season context.
- Added a fallback loop in `getClientIfExists` that normalizes base URLs to successfully look up client configurations by base URL when UUID lookups return `null`.
- Tracked active stream resolver/play session client via `_currentActiveResolverClient` and registered/restored active resolvers and player services during playback configuration in `_ensureResolverForItem`.
## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code

## Testing

- [X] Tested on physical device - tested on Android Mobile and Android TV

### Test Steps
1. Navigate to a show details page from the home row ("Recent TV Shows" or "Continue Watching").
2. Verify that the details page displays "Play" (instead of "Resume from S1:E1") for new, never watched shows.
3. Verify that the details page correctly displays the next unwatched episode (e.g., "Resume from S1:E4") for partially watched shows.
4. Play an episode from the "Continue Watching" details page and verify that the next episode is queued properly.

N/A

## Checklist

### DIFF
--- a/lib/data/services/media_server_client_factory.dart
+++ b/lib/data/services/media_server_client_factory.dart
@@ -27,7 +27,18 @@ class MediaServerClientFactory {
   }
 
   MediaServerClient? getClientIfExists(String serverId) {
-    return _clients[serverId];
+    final client = _clients[serverId];
+    if (client != null) return client;
+
+    final normalizedInput = normalizeServerBaseUrl(serverId);
+    if (normalizedInput.isNotEmpty) {
+      for (final activeClient in _clients.values) {
+        if (normalizeServerBaseUrl(activeClient.baseUrl) == normalizedInput) {
+          return activeClient;
+        }
+      }
+    }
+    return null;
   }
 
   MediaServerClient getActiveClient() {
--- a/lib/data/services/media_server_client_factory.dart
+++ b/lib/data/services/media_server_client_factory.dart
@@ -30,11 +30,13 @@ class MediaServerClientFactory {
     final client = _clients[serverId];
     if (client != null) return client;
 
-    final normalizedInput = normalizeServerBaseUrl(serverId);
-    if (normalizedInput.isNotEmpty) {
-      for (final activeClient in _clients.values) {
-        if (normalizeServerBaseUrl(activeClient.baseUrl) == normalizedInput) {
-          return activeClient;
+    if (serverId.contains('://')) {
+      final normalizedInput = normalizeServerBaseUrl(serverId);
+      if (normalizedInput.isNotEmpty) {
+        for (final activeClient in _clients.values) {
+          if (normalizeServerBaseUrl(activeClient.baseUrl) == normalizedInput) {
+            return activeClient;
+          }
         }
       }
     }
--- a/lib/di/modules/playback_module.dart
+++ b/lib/di/modules/playback_module.dart
@@ -501,7 +501,15 @@ void registerPlaybackModule() {
   );
 }
 
+MediaServerClient? _currentActiveResolverClient;
+
 void setActiveStreamResolver(MediaServerClient client) {
+  if (identical(client, _currentActiveResolverClient) &&
+      _getIt.isRegistered<MediaStreamResolver>() &&
+      _getIt.isRegistered<PlayerService>()) {
+    return;
+  }
+
   if (_getIt.isRegistered<MediaStreamResolver>()) {
     _getIt.unregister<MediaStreamResolver>();
   }
@@ -529,12 +537,14 @@ void setActiveStreamResolver(MediaServerClient client) {
   final manager = _getIt<PlaybackManager>();
   manager.setResolver(resolver);
   manager.setPlayerService(service);
+
+  _currentActiveResolverClient = client;
 }
 
 Future<void> _ensureResolverForItem(dynamic item) async {
   if (item is! AggregatedItem) return;
   final factory = _getIt<MediaServerClientFactory>();
-  final client = factory.getClientIfExists(item.serverId);
-  if (client == null || identical(client, _getIt<MediaServerClient>())) return;
+  final client = factory.getClientIfExists(item.serverId) ?? _getIt<MediaServerClient>();
   setActiveStreamResolver(client);
 }
+
--- a/lib/di/modules/playback_module.dart
+++ b/lib/di/modules/playback_module.dart
@@ -503,6 +503,10 @@ void registerPlaybackModule() {
 
 MediaServerClient? _currentActiveResolverClient;
 
+void resetActiveStreamResolver() {
+  _currentActiveResolverClient = null;
+}
+
 void setActiveStreamResolver(MediaServerClient client) {
   if (identical(client, _currentActiveResolverClient) &&
       _getIt.isRegistered<MediaStreamResolver>() &&
@@ -544,7 +548,8 @@ void setActiveStreamResolver(MediaServerClient client) {
 Future<void> _ensureResolverForItem(dynamic item) async {
   if (item is! AggregatedItem) return;
   final factory = _getIt<MediaServerClientFactory>();
-  final client = factory.getClientIfExists(item.serverId) ?? _getIt<MediaServerClient>();
+  final client = factory.getClientIfExists(item.serverId);
+  if (client == null) return;
   setActiveStreamResolver(client);
 }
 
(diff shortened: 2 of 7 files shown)


mattsigal:
Potentially related but currently unresolved issue - when playback is resumed from the Recent Movies or Recent Shows row (but not CW&NU) the play button doesn't seem to register properly. Flag for follow-up.

lib/data/services/media_server_client_factory.dart
```diff
@@ -27,7 +27,18 @@ class MediaServerClientFactory {
   }
 
   MediaServerClient? getClientIfExists(String serverId) {
-    return _clients[serverId];
+    final client = _clients[serverId];
+    if (client != null) return client;
+
+    final normalizedInput = normalizeServerBaseUrl(serverId);
```
RadicalMuffinMan: lines 33-40 replace with:
```
if (serverId.contains('://')) {
  final normalizedInput = normalizeServerBaseUrl(serverId);
  if (normalizedInput.isNotEmpty) {
    for (final activeClient in _clients.values) {
      if (normalizeServerBaseUrl(activeClient.baseUrl) == normalizedInput) {
        return activeClient;
      }
    }
  }
}
return null;
```

In lines 33-40 the new fallback loop is supposed to handle the case where a URL is passed instead of a server ID. But normalizeServerBaseUrl also returns a non empty string for regular server IDs so the loop runs and searches all clients on every failed lookup even when there's no URL involved. Add if (serverId.contains('://')) as a guard before the normalization block so the loop only runs when the input actually looks like a URL.

lib/di/modules/playback_module.dart
```diff
@@ -529,12 +537,14 @@ void setActiveStreamResolver(MediaServerClient client) {
...
+
+  _currentActiveResolverClient = client;
 }
 
 Future<void> _ensureResolverForItem(dynamic item) async {
   if (item is! AggregatedItem) return;
   final factory = _getIt<MediaServerClientFactory>();
-  final client = factory.getClientIfExists(item.serverId);
-  if (client == null || identical(client, _getIt<MediaServerClient>())) return;
+  final client = factory.getClientIfExists(item.serverId) ?? _getIt<MediaServerClient>();
```
RadicalMuffinMan:
```suggestion
  final client = factory.getClientIfExists(item.serverId);
  if (client == null) return;
```

if the server lookup returns null, the code now falls back to ?? _getIt<MediaServerClient>() (whatever server is currently active). On a single server setup this is fine, but on multi-server it would silently set up playback against the wrong server. The old behavior, just returning early when the server isn't found, was safer. Restore if (client == null) return; and drop the ?? fallback.

lib/di/modules/playback_module.dart
```diff
@@ -501,7 +501,15 @@ void registerPlaybackModule() {
   );
 }
 
+MediaServerClient? _currentActiveResolverClient;
```
RadicalMuffinMan, line 504:
```suggestion
MediaServerClient? _currentActiveResolverClient;

void resetActiveStreamResolver() {
  _currentActiveResolverClient = null;
}
```

_currentActiveResolverClient is a global variable and is never reset when the user logs out. If the same client object gets reused across sessions, setActiveStreamResolver will think it already did the setup and skip it. Add a small resetActiveStreamResolver() function that sets it back to null, and call it wherever the app tears down the session (same place disposeAll() is called on the client factory).

lib/ui/screens/detail/item_detail_screen.dart
```diff
@@ -4562,7 +4562,23 @@ class _ActionButtonsState extends State<_ActionButtons> {
       // Consume autoPlay from the current detail route so back navigation
       // doesn't re-trigger playback.
       context.replace(Destinations.item(item.id, serverId: item.serverId));
-      _play(context, item);
+
+      final isPhoto = item.type == 'Photo';
```
RadicalMuffinMan, line 4566: The same logic block is copy-pasted three times in item_detail_screen.dart
The chunk that computes totalEpisodes, isFullyWatched, isFullyUnwatched, isPartiallyWatched, and hasProgress now appears at lines 4566, 4636, and 5572. If the logic ever needs to change, all three copies need to be updated. Pull it into a private helper method (for example _computeWatchState(AggregatedItem)) and call that from all three places.

lib/ui/screens/detail/item_detail_screen.dart
```diff
@@ -4665,7 +4683,15 @@ class _ActionButtonsState extends State<_ActionButtons> {
           final s = nextUp.parentIndexNumber;
           final e = nextUp.indexNumber;
           if (s != null && e != null) {
```
RadicalMuffinMan, line 4691: lines 4685-4694 replace with:
```
if (s != null && e != null) {
  if (s == 1 && e == 1) {
    playButtonLabel = l10n.play;
  } else if (_isCompact(context)) {
    playButtonLabel = 'S${s}E$e';
  } else {
    playButtonLabel = 'Resume from S$s:E$e';
  }
}
```

In lines 4686-4694, the s == 1 && e == 1 check that shows "Play" instead of "Resume from S1:E1" is inside the _isCompact branch, so TV and desktop screens don't get the fix. Move the s == 1 && e == 1 check above the compact check so it applies everywhere.

packages/server_jellyfin/lib/src/api/jellyfin_item_fields.dart
```diff
@@ -1,2 +1,2 @@
 const kItemFields =
-    'Trickplay,Chapters,MediaSources,MediaStreams,People,Overview,Genres';
+    'Trickplay,Chapters,MediaSources,MediaStreams,People,Overview,Genres,RecursiveItemCount,ChildCount,SpecialFeatureCount';
```
RadicalMuffinMan: SpecialFeatureCount was added to the fields list but nothing in the new code actually reads it. RecursiveItemCount and ChildCount are both used but SpecialFeatureCount just adds to the response payload for no benefit. Remove it unless something else in the codebase already depends on it.

RadicalMuffinMan (changes requested):

mattsigal:
All comments addressed!

PR #429, opened 2026-06-08
fix(playback): resolve race condition to prevent OSD takeover

mattsigal:
# resolve race condition to prevent OSD takeover

## Summary
Resolves a race condition on desktop platforms using `media_kit` where the play/pause button state gets stuck on the Play icon and the OSD controls fail to auto-hide when video playback begins.

## Related Issues
None - bug found during playback testing

- Closes #
- Fixes #
- Related to #

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
- Implemented the `_mergeWithStale<T>` helper method in `MediaKitPlayerBackend` which merges the underlying streams with `_player.stream.playlist` events.
- Updated `playingStream` and `bufferingStream` getters to utilize this helper to ensure they are re-evaluated and re-emitted once `_isStale` becomes `false` when media loads.

## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code (but only impact Desktop clients)

## Testing

### Test Steps
1. Build and run the Windows application.
2. Play any video.
3. Verify that the play button switches to the **Pause** icon once playback begins.
4. Verify that the OSD correctly auto-hides after the inactivity delay.

## Checklist

### DIFF
--- a/lib/playback/media_kit_player_backend.dart
+++ b/lib/playback/media_kit_player_backend.dart
@@ -1125,16 +1125,16 @@ class MediaKitPlayerBackend implements PlayerBackend {
   });
 
   @override
-  Stream<bool> get playingStream => _player.stream.playing.map((playing) {
-    _updateStaleState();
-    return _isStale ? false : playing;
-  });
+  Stream<bool> get playingStream => _mergeWithStale<bool>(
+        _player.stream.playing,
+        () => _isStale ? false : _player.state.playing,
+      );
 
   @override
-  Stream<bool> get bufferingStream => _player.stream.buffering.map((buffering) {
-    _updateStaleState();
-    return _isStale ? false : buffering;
-  });
+  Stream<bool> get bufferingStream => _mergeWithStale<bool>(
+        _player.stream.buffering,
+        () => _isStale ? false : _player.state.buffering,
+      );
 
   @override
   Stream<bool> get completedStream => _player.stream.completed.map((completed) {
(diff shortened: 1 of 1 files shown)


lib/playback/media_kit_player_backend.dart
```diff
@@ -1125,16 +1125,16 @@ class MediaKitPlayerBackend implements PlayerBackend {
   });
 
   @override
```
RadicalMuffinMan, line 1137:
```suggestion

```

lib/playback/media_kit_player_backend.dart
```diff
@@ -1125,16 +1125,16 @@ class MediaKitPlayerBackend implements PlayerBackend {
...
-  Stream<bool> get playingStream => _player.stream.playing.map((playing) {
-    _updateStaleState();
-    return _isStale ? false : playing;
-  });
+  Stream<bool> get playingStream => _mergeWithStale<bool>(
```
RadicalMuffinMan:
```suggestion
   late final Stream<bool> _playingStream = _mergeWithStale<bool>(
```

lib/playback/media_kit_player_backend.dart
```diff
@@ -1125,16 +1125,16 @@ class MediaKitPlayerBackend implements PlayerBackend {
...
+        _player.stream.playing,
+        () => _isStale ? false : _player.state.playing,
+      );
 
   @override
-  Stream<bool> get bufferingStream => _player.stream.buffering.map((buffering) {
-    _updateStaleState();
-    return _isStale ? false : buffering;
-  });
+  Stream<bool> get bufferingStream => _mergeWithStale<bool>(
```
RadicalMuffinMan:
```suggestion
  @override
  Stream<bool> get playingStream => _playingStream;

  late final Stream<bool> _bufferingStream = _mergeWithStale<bool>(
```

lib/playback/media_kit_player_backend.dart
```diff
@@ -1125,16 +1125,16 @@ class MediaKitPlayerBackend implements PlayerBackend {
...
+        _player.stream.buffering,
+        () => _isStale ? false : _player.state.buffering,
+      );
```
RadicalMuffinMan, line 1142:
```suggestion

@override
  Stream<bool> get bufferingStream => _bufferingStream;
```

RadicalMuffinMan (commented):
A getter like get playingStream => _mergeWithStale(...) re-runs its right-hand side every single time something reads playingStream. So each read builds a fresh StreamController and wires up its own two subscriptions to the player. Today the code reads it only once, so you never notice. But the moment two places read playingStream, you end up with two separate controllers both hooked into the same player, and the spare one just sits there listening forever because nothing ever tears it down.

late final is Dart's way of saying "build this once, then keep handing back the same thing." The first time the field gets read, Dart runs the initializer (the _mergeWithStale(...) call), saves the result, and from then on every read returns that saved value. So no matter how many times anyone touches playingStream, the stream is built a single time.

The getter now just returns that saved field (get playingStream => _playingStream), so nothing changes for the people using it. Callers still write backend.playingStream.listen(...) exactly like before.

mattsigal:
Fixes incorporated, thanks!

PR #436, opened 2026-06-09
fix(auth): Fix Auto-Logon "Current User" behaviour

mattsigal:
# Fix and refine the "Current User" login behaviour

## Summary
The bug: the "Current User" setting was not functioning as intended (it was always triggering the user selection screen instead of auto logging in).

The fix: Refined the "Auto Login" setting logic for "Current User" so that the auto-login user target is kept persistent on session switch and sign out, only updating when explicitly changed in the Settings menu. This prevents child/other accounts from automatically overwriting the auto-login configuration on a shared device. Also improved the UX by introducing a `dialogLabelOf` callback to `EnumPreferenceTile` to ensure the picker options show the clear, localized option "Current User", while the main settings screen correctly displays the name of the designated auto-login user (e.g., "Matt") when viewed from other accounts. This can be overridden by the other accounts (by switching auto log-in to disable and then turning Current User back on) if desired.

## Related Issues
Bug found during testing.

- Closes #
- Fixes #
- Related to #

## Type of Change
- [X] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Performance improvement
- [X] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
- Added `dialogLabelOf` property to `EnumPreferenceTile` in `preference_tiles.dart` to allow displaying different labels in the picker dialog versus the summary tile.
- Configured `dialogLabelOf` on the Auto Login preference tile in `settings_side_panel.dart` to return the localized `currentUser` inside the choice picker dialog, while keeping the summary label showing the auto-login user's username on other accounts.
- Removed the automatic call to `_authPrefs.setAutoLogin` in `SessionRepository.switchCurrentSession` to prevent switching accounts from overwriting the configured auto-login target.
- Modified `destroyCurrentSession` in `session_repository.dart` to avoid clearing the saved auto-login credentials when signing out if behavior is set to `currentUser`.

## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code

## Testing

### Test Steps
1. Log in to a main account (e.g., Matt) and set "Auto Login" to "Current User" in settings.
2. Close and restart the app to verify it auto-logs into the main account.
3. Log in to a second account (e.g., child account) and check settings.
4. Verify that the "Auto Login" tile summary correctly shows the main account's name (e.g. "Matt") instead of "Current User".
5. Click the tile to open the picker dialog and verify that the option is labeled "Current User" (not the main account name).
6. Restart the app from the second account, verifying it still auto-logs back into the main account.

The setting in question:
2026-06-08_16-38-11_moonfin

Fixed version, my account with auto login set to "Current User"
2026-06-08_17-35-42_moonfin

Fixed version, other account when auto login is set to "Matt"
2026-06-08_17-35-56_moonfin

## Checklist

### DIFF
--- a/lib/ui/screens/settings/settings_side_panel.dart
+++ b/lib/ui/screens/settings/settings_side_panel.dart
@@ -24,6 +24,8 @@ import '../../../util/app_distribution.dart';
 import '../../widgets/app_update_dialog.dart';
 
 import '../../../auth/store/authentication_preferences.dart';
+import '../../../auth/store/authentication_store.dart';
+import '../../../auth/repositories/session_repository.dart';
 import '../../../l10n/app_localizations.dart';
 import '../../../playback/audio_capability_profile.dart';
 import '../../../playback/external_player_service.dart';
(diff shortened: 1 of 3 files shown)


lib/ui/screens/settings/settings_side_panel.dart
RadicalMuffinMan, line 1: Right now the settings screen is doing login bookkeeping itself. Inside the auto-login tile it reaches into the auth system, figures out who the auto-login user is, looks up their name, and saves/clears the auto-login target. That's login logic living in a UI file. The rule of thumb: screens should ask for things and display things; the actual "save who auto-logs in" work should live in SessionRepository, which already owns sessions and accounts. Keeping it there means it's reusable, testable, and there's one obvious place to look when login behavior changes.

add three small helpers to SessionRepository (lib/auth/repositories/session_repository.dart), right after the activeUserId getter:
```
  // save or clear the auto login target based on the chosen behavior.
  Future<void> applyAutoLoginForBehavior(UserSelectBehavior behavior) async {
    if (behavior == UserSelectBehavior.currentUser) {
      final serverId = _activeServerId;
      final userId = _activeUserId;
      if (serverId != null && userId != null) {
        await _authPrefs.setAutoLogin(serverId, userId);
      }
    } else {
      await _authPrefs.clearAutoLogin();
    }
  }

  // true when the person using the app right now is the saved auto login user.
  bool get activeUserIsAutoLoginTarget =>
      _activeUserId != null &&
      _activeServerId != null &&
      _authPrefs.savedAutoLoginUserId == _activeUserId &&
      _authPrefs.savedAutoLoginServerId == _activeServerId;

  // the name of the saved auto-login user or null if none.
  String? autoLoginTargetDisplayName() {
    final serverId = _authPrefs.savedAutoLoginServerId;
    final userId = _authPrefs.savedAutoLoginUserId;
    if (serverId.isEmpty || userId.isEmpty) return null;
    return _authStore.getUser(serverId, userId)?.name;
  }
```

Everything these use (_authPrefs, _authStore, _activeServerId, _activeUserId) already exists in that class, so you're just adding methods, not new dependencies.


Then make the settings tile thin (lib/ui/screens/settings/settings_side_panel.dart). Replace the big labelOf block (lines 473-498) and onChanged block (lines 504-518) with these short versions that just call the helpers:
```
            labelOf: (v) {
              if (v == UserSelectBehavior.currentUser) {
                final session = GetIt.instance<SessionRepository>();
                // If I'm the auto-login user, just say "Current User"
                // If someone else is, show their name so it's clear
                if (session.activeUserIsAutoLoginTarget) return l10n.currentUser;
                return session.autoLoginTargetDisplayName() ?? l10n.currentUser;
              }
              return switch (v) {
                UserSelectBehavior.disabled => l10n.disabled,
                UserSelectBehavior.lastUser => l10n.lastUser,
                _ => l10n.currentUser,
              };
            },
            dialogLabelOf: (v) => switch (v) {
              UserSelectBehavior.disabled => l10n.disabled,
              UserSelectBehavior.lastUser => l10n.lastUser,
              UserSelectBehavior.currentUser => l10n.currentUser,
            },
            onChanged: () {
              final behavior = GetIt.instance<PreferenceStore>()
                  .get(UserPreferences.autoLoginUserBehavior);
              GetIt.instance<SessionRepository>()
                  .applyAutoLoginForBehavior(behavior);
            },
```

lib/ui/screens/settings/settings_side_panel.dart
```diff
@@ -24,6 +24,8 @@ import '../../../util/app_distribution.dart';
 import '../../widgets/app_update_dialog.dart';
 
 import '../../../auth/store/authentication_preferences.dart';
+import '../../../auth/store/authentication_store.dart';
```
RadicalMuffinMan:
```suggestion
```

RadicalMuffinMan (changes requested):

mattsigal:
Good suggestions! All incorporated.

PR #467, opened 2026-06-10
fix: stop audio from switching to match subtitles

### LINKED ISSUE #428: [UI] Automatic audio setting is not reflected in UI during playback
### Moonfin version

2.0.1

### Platform

Android TV

### OS version

Android 14

### Device model

Sony BRAVIA 4K VH2

### Where does it happen?

Player UI (controls/OSD)

### Input method

D-pad / remote

### Steps to reproduce

Hello! It's me again. This issue is about Linux Desktop + Android TV. It sounds like a feature request but I'm targeting the UI issue.

This issue solved a very particular problem #45 but has one flaw. As many people, I prefer to watch anime in Japanese audio, but anything else I play it on my default language (latin american spanish). My OS language on my android TV and Desktop is latin spanish and Moonfin correctly picks it up and sets latin spanish audio for all my shows as #45 expected.

Moonfin picks my System Language and correctly plays the video with my System Language's audio. But it isn't updating in the UI and still displays the audio language I was using yesterday before opening up Moonfin.

The UI says "Japanese", but the audio is actually spanish.
Image

I'm requesting that the audio dropdown be updated when Moonfin changes the audio track.

My workaround, as I prefer japanese, involves changing audio 2 times. First changing to English, then changing back to Japanese.

### Screenshots / recordings (optional)

_No response_

mattsigal:
# PR: The one about calming down the language streams

## Summary
Fix the Audio Track Swap bug where changing subtitles to the system language (e.g. Spanish) causes the Jellyfin server to silently swap the audio track because the active audio stream index was unresolved (`null`).

## Related Issues

- Closes #
- Fixes #428 
- Related to #

## Type of Change
- [X ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
- In `_playCurrentItem()` (after `_currentResolution = resolution` is set), if `_audioStreamIndex` is `null`, we locate the audio stream marked as selected by the server (via `IsSelected` or `isSelected` property in `resolution.mediaStreams`), and assign its index to `_audioStreamIndex`. If not explicitly marked as selected, we fallback to the default audio stream (`IsDefault`), or the first available audio stream.
- Similarly, if `_subtitleStreamIndex` is `null`, we extract the selected subtitle stream index or default to `-1` (disabled).
- Reset `_audioStreamIndex` and `_subtitleStreamIndex` to `null` in `playOffline()`.

## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X ] All / Shared code

## Testing

- [X ] Manual testing completed
- [] Not tested (explain why): 

### Test Steps
1. Play an episode with no explicit audio selection (audio resolving automatically to Japanese).
2. Verify that `_audioStreamIndex` is populated with the resolved Japanese index.
3. Switch subtitles (which triggers re-resolution).
4. Verify that the request explicitly requests the Japanese audio track index and the audio remains in Japanese (is not swapped to the system language).

## Checklist
- [X ] Code builds successfully
- [X ] Code follows project style and conventions
- [X ] No unnecessary commented-out code
- [X ] No new warnings introduced

### DIFF
--- a/packages/playback_core/lib/src/playback_manager.dart
+++ b/packages/playback_core/lib/src/playback_manager.dart
@@ -891,6 +891,18 @@ class PlaybackManager implements AudioOwnable {
     _mediaSourceId = resolution.mediaSourceId;
     _itemKnownDuration = _resolvedItemDuration(item, resolution.mediaSourceId);
 
+    if (_audioStreamIndex == null) {
+      final audioStreams =
+          resolution.mediaStreams.where((s) => s['Type'] == 'Audio').toList();
+      if (audioStreams.isNotEmpty) {
+        final defaultAudio = audioStreams.firstWhere(
+          (s) => s['IsDefault'] == true,
+          orElse: () => audioStreams.first,
+        );
+        _audioStreamIndex = defaultAudio['Index'] as int?;
+      }
+    }
+
     final playbackDecisionLogger = _playbackDecisionLogger;
     if (playbackDecisionLogger != null && _backend != null) {
       try {
@@ -1816,6 +1828,8 @@ class PlaybackManager implements AudioOwnable {
     _lastExplicitAudioLanguage = null;
     _lastExplicitSubtitleLanguage = null;
     _lastExplicitSubtitleEnabled = null;
+    _audioStreamIndex = null;
+    _subtitleStreamIndex = null;
     _isAutoNexting = false;
     _isManualNexting = false;
     suppressAutoNext = false;


RadicalMuffinMan:
rebase to main

mattsigal:
rebased

packages/playback_core/lib/src/playback_manager.dart
```diff
@@ -891,6 +891,49 @@ class PlaybackManager implements AudioOwnable {
     _mediaSourceId = resolution.mediaSourceId;
     _itemKnownDuration = _resolvedItemDuration(item, resolution.mediaSourceId);
 
+    if (_audioStreamIndex == null) {
```
RadicalMuffinMan, line 894:
```
    if (_audioStreamIndex == null) {
      final audioStreams =
          resolution.mediaStreams.where((s) => s['Type'] == 'Audio').toList();
      if (audioStreams.isNotEmpty) {
        final defaultAudio = audioStreams.firstWhere(
          (s) => s['IsDefault'] == true,
          orElse: () => audioStreams.first,
        );
        _audioStreamIndex = defaultAudio['Index'] as int?;
      }
    }
```
The first time we play an item we send no audio index, so the server picks one for us. Remember that pick here. Otherwise the next time we re-resolve (for example when the user turns subtitles on) we again send a null audio index, and the server is free to swap the audio track to match the new subtitle language. Pinning the index keeps audio stable.

The PR code checks IsSelected/isSelected, but Jellyfin and Emby never put that field on a stream, so those branches never run and only the IsDefault path was ever doing the work. Dropping the dead checks makes the code say what it actually does. Second, the code block repeats .where((s) => s['Type'] == 'Audio') three times; computing the list once is shorter and easier to follow.

packages/playback_core/lib/src/playback_manager.dart
```diff
@@ -891,6 +891,49 @@ class PlaybackManager implements AudioOwnable {
...
+              .where((s) => s['Type'] == 'Audio')
+              .firstOrNull;
+          if (firstAudio != null) {
+            _audioStreamIndex = firstAudio['Index'] as int?;
+          }
+        }
+      }
+    }
+
+    if (_subtitleStreamIndex == null) {
```
RadicalMuffinMan: remove the whole `if (_subtitleStreamIndex == null)` block. The bug is about audio swapping. When the user changes a subtitle, the app already sets a real subtitle index before re-resolving, so this block never runs on the path the code is fixing. The only time it does run is the very first play, where it forces the subtitle index to -1 (off).

RadicalMuffinMan (changes requested):
also rebase on main, i think the changes from pr #466 might get overwritten otherwise

mattsigal:
Rebased and updated

PR #498, opened 2026-06-11
fix: resolve folder navigation and image display bugs in nested mixed libraries

### LINKED ISSUE #262: [UI] Mixed movies and shows folders don't display correctly
### Moonfin version

1.4.0

### Platform

Linux

### OS version

Nobara Linux 43

### Device model

Personal PC

### Where does it happen?

Library

### Input method

Mouse / trackpad

### Steps to reproduce

I've found that the mixed movies and shows library has some issues.  I've noticed that it will always sort series first and then movies regardless of sort title. I have a library that sorts them in a watch order so the series and movies should be interspersed hence the sort order.

Another, arguably larger , issue I've seen with the same library type is that  when you have the same mixed movies and tv shows library if the library is nested at all the display gets super off. I have a library called Caetgories which is a folder that then has themed libraries nested inside, so as an example I have Categories > Marvel > MCU and then the last folder has all the content. Jellyfin's native clients display this just fine however Moonfin will interpret the series as further subfolders instead of series despite them being marked as "Series" in Jellyfin and on top of that all the movies ignore the poster and try to display the thumb inside the poster dimensions. Removing the thumb forces it to display the poster instead but if it has a thumb it defaults to that regardless of the poster type. Also the sort by and display settings are not available after going more than one folder deep. So If you have Library > sub folder > content these display issues start whereas Library only does display fine (Other than putting all the series together before the movies

I can also reproduce this on my android tv client as well so it does not appear to be client specific from my testing

### Screenshots / recordings (optional)

_No response_

mattsigal:
# The one about Mixed Media

## Summary
Resolves mixed content library nested folder browsing issues where Series/Seasons/Collections/Playlists are incorrectly navigated as folders and media items use landscape thumbs instead of portrait posters.

## Related Issues

- Closes #
- Fixes #262 
- Related to #

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
- Modified `FolderBrowseViewModel.isNavigableFolder` to return `false` for `Series`, `Season`, `BoxSet`, and `Playlist`.
- Modified `FolderBrowseScreen._imageUrl` to check folder navigation status. Prioritizes `Primary` poster images over `Thumb` landscape images for media items, and keeps `Thumb` prioritization for actual folders.
- Fixed outdated assertions checking `pref_audio_preference_split_v1` in `audio_migration_test.dart` (updated to `v2`) so that all tests pass.

## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code

## Testing

### Test Steps
1. Navigate into a nested folder containing movies and TV shows.
2. Verify that movies and TV shows display using their portrait primary posters.
3. Verify that clicking on a TV series navigates to the series detail screen rather than folder view.

## Checklist

### DIFF
--- a/lib/data/viewmodels/folder_browse_view_model.dart
+++ b/lib/data/viewmodels/folder_browse_view_model.dart
@@ -208,16 +208,23 @@ class FolderBrowseViewModel extends ChangeNotifier {
 
   bool isNavigableFolder(AggregatedItem item) {
     final type = item.type;
+    if (type == 'Series' ||
+        type == 'Season' ||
+        type == 'BoxSet' ||
+        type == 'Playlist') {
+      return false;
+    }
+
+    final isFolder = item.rawData['IsFolder'] as bool? ?? false;
+    if (isFolder) return true;
+
     return type == 'Folder' ||
         type == 'CollectionFolder' ||
         type == 'UserView' ||
-        type == 'BoxSet' ||
         type == 'MusicAlbum' ||
-        type == 'Season' ||
-        type == 'Series' ||
-        type == 'PhotoAlbum' ||
-        type == 'Playlist';
+        type == 'PhotoAlbum';
   }
+
 
   @override
   void dispose() {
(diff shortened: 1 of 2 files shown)


RadicalMuffinMan:
rebase on main

PR #515, opened 2026-06-12
UI - Clean up Home Row Toggles behaviour

mattsigal:
# Pull Request

## Summary
Cleans up the Home Row Toggles settings screen UI on Android TV by highlighting only the list icon button inside when focused on TV, fixing sticky header visual glitches when scrolling quickly back to the top, and preventing double navigation pushes.

## Related Issues
Part of BrokenDroid's list of UI improvements.

- Closes #
- Fixes #
- Related to #

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [X] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
* The inner list icon button is wrapped in a Focus widget, making it the only focusable element in the row, rendering a secondary accent focus highlight and shadow when focused on TV.
* Set scrolledUnderElevation: 0 inside buildSettingsAppBar in settings_app_bar.dart and on appBarTheme globally in app_theme.dart. This disables Material 3 scrolled-under elevation changes on settings screen headers, ensuring they remain flat and match the background color without getting stuck during rapid scrolls.
* Updated pushSettingsScreen inside settings_panel.dart to return a Future<void>. Implemented a _navigating guard boolean in home_row_toggles_screen.dart that locks navigation during transition to prevent pushing the Home Sections Screen twice.

## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code

## Testing

- [X] Manual testing completed - tested on the Shield.

### Test Steps
1. Navigate to Settings -> Personalization -> Home Row Toggles.
2. Verify that the top description tile container renders with a matching unfocused outline border.
3. Verify that only the list icon button inside the description tile is focusable on TV, rendering a secondary focus color glow/border (cyan on Neon Pulse) while the icon design color remains pink.
4. Scroll quickly to the bottom and back to the top. Verify that the app bar stays flat and does not show any lingering scroll-under colors or borders.
5. Double-click or click rapidly on the top description tile's button. Verify that the Home Sections Screen opens exactly once.

All behavior changes.

## Checklist

### DIFF
--- a/lib/ui/screens/settings/home_row_toggles_screen.dart
+++ b/lib/ui/screens/settings/home_row_toggles_screen.dart
@@ -182,19 +182,19 @@ class _HomeRowTogglesScreenState extends State<HomeRowTogglesScreen> {
                       child: Container(
                         decoration: BoxDecoration(
                           color: _buttonFocused
-                              ? theme.colorScheme.primary.withValues(alpha: 0.3)
+                              ? AppColorScheme.accent.withValues(alpha: 0.18)
                               : theme.colorScheme.primary.withValues(alpha: 0.15),
                           shape: BoxShape.circle,
                           border: Border.all(
                             color: _buttonFocused
-                                ? theme.colorScheme.primary
+                                ? AppColorScheme.accent
                                 : theme.colorScheme.primary.withValues(alpha: 0.35),
                             width: 1.5,
                           ),
                           boxShadow: _buttonFocused
                               ? [
                                   BoxShadow(
-                                    color: theme.colorScheme.primary.withValues(alpha: 0.22),
+                                    color: AppColorScheme.accent.withValues(alpha: 0.22),
                                     blurRadius: 14,
                                     spreadRadius: 0.5,
                                   ),
--- a/lib/ui/screens/settings/home_row_toggles_screen.dart
+++ b/lib/ui/screens/settings/home_row_toggles_screen.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/material.dart';
 import 'package:get_it/get_it.dart';
 import 'package:server_core/server_core.dart';
+import 'package:moonfin_design/moonfin_design.dart';
 
 import '../../../data/services/plugin_sync_service.dart';
 import '../home/home_view_model.dart';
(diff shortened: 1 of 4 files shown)


lib/ui/screens/settings/home_row_toggles_screen.dart
```diff
@@ -150,20 +177,37 @@ class _HomeRowTogglesScreenState extends State<HomeRowTogglesScreen> {
...
                     ),
                     const SizedBox(width: 10),
-                    Container(
-                      decoration: BoxDecoration(
-                        color: theme.colorScheme.primary.withValues(alpha: 0.15),
-                        shape: BoxShape.circle,
-                        border: Border.all(
-                          color: theme.colorScheme.primary.withValues(alpha: 0.35),
-                          width: 1.5,
+                    Focus(
```
RadicalMuffinMan, line 180:
```suggestion
                    Focus(
                      canRequestFocus: false,
                      skipTraversal: true,
```

 A plain Focus widget is itself a focusable element. The IconButton inside it is also focusable. So right now there are two focusable things stacked in the same spot, which on a D-pad usually means an extra, useless focus stop (you might have to press the remote twice to move past the button). Adding canRequestFocus: false tells the wrapper "you are not a focus target yourself," and skipTraversal: true keeps it out of the D-pad navigation order. The onFocusChange callback still fires correctly, because a Focus node reports hasFocus = true whenever any descendant (here, the IconButton) is focused so _buttonFocused keeps working exactly as before.

lib/ui/theme/app_theme.dart
```diff
@@ -78,6 +78,7 @@ class AppTheme {
       appBarTheme: AppBarTheme(
         backgroundColor: c.background,
         elevation: 0,
+        scrolledUnderElevation: 0,
```
RadicalMuffinMan:
```suggestion
        
```
This line is on the global appBarTheme, so it changes the app bar on every screen in the app (home, library, detail, live TV, etc), not just settings. That's a much bigger blast radius than a "Home Row Toggles cleanup"

RadicalMuffinMan (changes requested):

lib/ui/screens/settings/home_row_toggles_screen.dart
mattsigal: Done! Added \canRequestFocus: false\ and \skipTraversal: true\ so the D-pad traverses directly to the focusable \IconButton\ without hitting the outer \Focus\ container.

lib/ui/theme/app_theme.dart
mattsigal: Done! Reverted this change in the global theme so it does not affect non-settings screen AppBars. The settings-specific header elevation fix is now correctly isolated to buildSettingsAppBar in settings_app_bar.dart.

PR #516, opened 2026-06-12
feat: clean up Seerr config screen

mattsigal:
# Pull Request

## Summary
No text/localization changes but a bunch of under the hood finessing. Cleans up the Settings -> Integrations -> Seerr configuration screen layout, focus traversal behavior, and reordering logic on Android TV/mobile. This ensures visual consistency with Rating Sources config.

## Related Issues
Droid's list of UI issues.

- Closes #
- Fixes #
- Related to #

## Type of Change
- [X] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Performance improvement
- [X] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
1. Fixed Initial Focus & D-pad Navigation Links to follow up/down logic.
2. Removed manual arrowDown overriding logic from "Sign In" and "Sign Out" buttons to allow Flutter's default 2D directional navigation traversal to land on the NSFW Filter toggle immediately.
3. Prevented the _SeerrLoginCard from stealing focus on the initial asynchronous load unless focus is already within the card.
4. Redesigned the Sign Out and Sign In buttons layout: wrapped the inner 5. AnimatedContainer in the Focus node so that the container draws a clean outer Cyan glow shadow (AppColorScheme.onSurface), while the FilledButton itself draws its Cyan border exactly along its outline (side: isFocused ? BorderSide(...) : BorderSide.none), matching the button dimensions.
5. Removed explicit scroll alignment callbacks (_ensureFocusVisible) from "Sign In" and "Sign Out" buttons to prevent jarring scroll jumps between adjacent settings elements.
6.. Re-styled _SeerrReorderableTile using the exact layout pattern of _ReorderableTile in ratings_config_screen.dart. Replaced card boundaries, borders, margins, shadows, and text color overrides with a flat background layout style where focused rows highlight with colorScheme.primary.withValues(alpha: 0.18) and unfocused rows are transparent, matching the standard full-width settings menu items.
7. Restructured row focus by attaching the focus node to the outer Focus widget of the row tile rather than the inner ListTile.
8. Prevented the page from clearing and rebuilding all row FocusNodes on every sync notification by checking if the row size has actually changed. This preserves focus on the active moving item during D-pad LEFT/RIGHT reordering.

## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code

## Testing

- [X] Manual testing completed - tested on Shield

### Test Steps
1. Navigate to Settings -> Integrations -> Seerr. Focus lands on "Enable Seerr".
2. Press UP to go to the Refresh button. Press DOWN to return to "Enable Seerr".
3. Press DOWN to go to the Seerr Account Type card (Jellyfin/local auth option, or Sign Out).
4. Verify the Sign Out button draws a themed Cyan outline and glow.
5. Press DOWN from the Sign Out button. Verify it navigates to the NSFW Filter toggle instantly without scrolling the viewport jarringly.
6.Check or uncheck a Discover Row. Press LEFT to move it up or RIGHT to move it down. Verify that focus stays pinned to the moving item as it transitions.

Shield_Screenshot_2026-06-12_15-27-25

Shield_Screenshot_2026-06-12_15-27-45

## Checklist

### DIFF
--- a/lib/ui/screens/settings/seerr_config_screen.dart
+++ b/lib/ui/screens/settings/seerr_config_screen.dart
@@ -584,6 +584,8 @@ class _SeerrLoginCardState extends State<_SeerrLoginCard> {
   final _passwordFocus = FocusNode(debugLabel: 'seerr_password');
   final _localAuthFocus = FocusNode(debugLabel: 'seerr_auth_local');
   final _signInFocus = FocusNode(debugLabel: 'seerr_sign_in');
+  final _signInButtonFocus = FocusNode(canRequestFocus: false, debugLabel: 'seerr_sign_in_btn');
+  final _signOutButtonFocus = FocusNode(canRequestFocus: false, debugLabel: 'seerr_sign_out_btn');
   final _usernameTvFieldKey = GlobalKey<CustomTVTextFieldState>();
   final _passwordTvFieldKey = GlobalKey<CustomTVTextFieldState>();
 
@@ -685,6 +687,8 @@ class _SeerrLoginCardState extends State<_SeerrLoginCard> {
     _passwordFocus.dispose();
     _localAuthFocus.dispose();
     _signInFocus.dispose();
+    _signInButtonFocus.dispose();
+    _signOutButtonFocus.dispose();
     super.dispose();
   }
 
@@ -1242,6 +1246,7 @@ class _SeerrLoginCardState extends State<_SeerrLoginCard> {
                                   : null,
                             ),
                             child: FilledButton(
+                              focusNode: PlatformDetection.isTV ? _signOutButtonFocus : null,
                               style: FilledButton.styleFrom(
                                 foregroundColor: colorScheme.onPrimary,
                                 backgroundColor: colorScheme.primary,
@@ -1387,6 +1392,7 @@ class _SeerrLoginCardState extends State<_SeerrLoginCard> {
                                   : null,
                             ),
                             child: FilledButton(
+                              focusNode: PlatformDetection.isTV ? _signInButtonFocus : null,
                               onPressed: _canSignIn ? _submitSignIn : null,
                               style: FilledButton.styleFrom(
                                 shape: RoundedRectangleBorder(
--- a/lib/ui/screens/settings/seerr_config_screen.dart
+++ b/lib/ui/screens/settings/seerr_config_screen.dart
@@ -15,7 +15,11 @@ import '../../../preference/seerr_row_config.dart';
 import '../../../preference/user_preferences.dart';
 import '../../../util/focus/dpad_keys.dart';
 import '../../../util/platform_detection.dart';
+import 'package:moonfin_design/moonfin_design.dart';
+
+import '../../widgets/focus/request_initial_focus.dart';
 import '../../widgets/settings/clean_settings_typography.dart';
+import '../../widgets/settings/preference_tiles.dart';
 import 'settings_app_bar.dart';
 
 typedef _SeerrSignInAction =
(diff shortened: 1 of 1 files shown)


lib/ui/screens/settings/seerr_config_screen.dart
```diff
@@ -471,17 +582,48 @@ class _SeerrLoginCardState extends State<_SeerrLoginCard> {
   final _passwordController = TextEditingController();
   final _usernameFocus = FocusNode(debugLabel: 'seerr_username');
   final _passwordFocus = FocusNode(debugLabel: 'seerr_password');
-  final _jellyfinAuthFocus = FocusNode(debugLabel: 'seerr_auth_jellyfin');
   final _localAuthFocus = FocusNode(debugLabel: 'seerr_auth_local');
   final _signInFocus = FocusNode(debugLabel: 'seerr_sign_in');
```
RadicalMuffinMan, line 586:
```suggestion
  final _signInFocus = FocusNode(debugLabel: 'seerr_sign_in');
  final _signInButtonFocus = FocusNode(canRequestFocus: false, debugLabel: 'seerr_sign_in_btn');
  final _signOutButtonFocus = FocusNode(canRequestFocus: false, debugLabel: 'seerr_sign_out_btn');
```

A FilledButton is focusable by default. Right now each button is Focus(...) --> FilledButton(...), so there are two focusable things stacked and since this PR switched these buttons to rely on Flutter's default D-pad traversal, the remote will likely stop on both (one wasted press). A FocusNode with canRequestFocus: false means "this node can never take focus," so the inner button drops out of the D-pad path and only the outer Focus remains. The outer Focus already handles the click (Select --> _submitSignIn/_submitSignOut), so nothing is lost.

lib/ui/screens/settings/seerr_config_screen.dart
```diff
@@ -530,10 +683,8 @@ class _SeerrLoginCardState extends State<_SeerrLoginCard> {
     _passwordController.dispose();
     _usernameFocus.dispose();
     _passwordFocus.dispose();
-    _jellyfinAuthFocus.dispose();
     _localAuthFocus.dispose();
     _signInFocus.dispose();
```
RadicalMuffinMan, line 689:
```suggestion
    _signInFocus.dispose();
    _signInButtonFocus.dispose();
    _signOutButtonFocus.dispose();
```
every FocusNode you create must be disposed, or it leaks.

lib/ui/screens/settings/seerr_config_screen.dart
```diff
@@ -1073,30 +1213,62 @@ class _SeerrLoginCardState extends State<_SeerrLoginCard> {
...
+                                  ? [
+                                      BoxShadow(
+                                        color: AppColorScheme.onSurface.withValues(alpha: 0.35),
+                                        blurRadius: 10,
+                                        spreadRadius: 1.5,
+                                      ),
+                                    ]
+                                  : null,
+                            ),
+                            child: FilledButton(
```
RadicalMuffinMan, line 1248:
```suggestion
                            child: FilledButton(
                               focusNode: PlatformDetection.isTV ? _signOutButtonFocus : null,
```

On TV we hand the button our non-focusable node, so the outer Focus is the single focus target (glow shows because _signInFocus.hasFocus is still true, and Select still submits via the outer handler). On phone/desktop we pass null, so the button keeps its normal focusable behavior for touch and keyboard/tab.

lib/ui/screens/settings/seerr_config_screen.dart
```diff
@@ -1192,26 +1365,53 @@ class _SeerrLoginCardState extends State<_SeerrLoginCard> {
...
```
RadicalMuffinMan, line 1394:
```suggestion
                            child: FilledButton(
                               focusNode: PlatformDetection.isTV ? _signOutButtonFocus : null,
```

On TV we hand the button our non-focusable node, so the outer Focus is the single focus target (glow shows because _signInFocus.hasFocus is still true, and Select still submits via the outer handler). On phone/desktop we pass null, so the button keeps its normal focusable behavior for touch and keyboard/tab.

RadicalMuffinMan (changes requested):

PR #539, opened 2026-06-15
fix(playback): persist media info selections, track propagation, and …

mattsigal:
# Pull Request

## Summary
This PR addresses details page layout state synchronization and how selected version/track details propagate to the player and subsequent queue episodes. It ensures that the selected version, audio, and subtitle tracks selected on the details screen are correctly initially passed to the player and propagate to subsequent queue episodes using robust ISO-639 language matching rules.

## Related Issues
* Playback queuing: audio/subtitle selection from media info ignored during playback unless explicitly re-selected.
* Custom track selections do not follow from one media item to the next in queue (e.g. Arabic subtitles reverting to English).

- Closes #
- Fixes #
- Related to #

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

### Playback Core Components
* Explicit Selection Tracking: Added fields _audioSelectionExplicit, _pendingItemAudioSelectionExplicit, and _pendingItemSubtitleSelectionExplicit to track whether track selections were made explicitly by the user or auto-detected.
* Propagation of Track Selections: Updated playItems() and setPendingItemOverrides() to accept and process these explicit flags.
* ISO-639 Compliant Language Matching & Propagation:
    * In _playCurrentItem(), after successful stream resolution, if track selections are marked as explicit, we look up the stream's language and store it in _lastExplicitAudioLanguage or _lastExplicitSubtitleLanguage/_lastExplicitSubtitleEnabled.
    * Added helpers _languagesMatch(), _normalizeLanguage(), and _toIso3() along with the full _kIso6391To6392 code map to normalize and compare language tags.
    * Updated _translateTrackSelectionsForNewItem() to compare languages using the _languagesMatch() helper. This resolves issues where language tags with different ISO formats (e.g., 2-letter ar vs 3-letter ara or full name Arabic) failed to propagate track selections across queue items.

### Title and DisplayTitle Fallback Parsing:
* Added _extractLanguage(Map? stream) to parse language information from the stream's Title and DisplayTitle when the server returns undetermined (und) or null for the language tag (common for external/side-loaded SRT subtitle tracks).
* Added a token-to-ISO-639-2 keywords map (_kLanguageKeywords) to map tokens like "Arabic" or "ara" case-insensitively to 'ara'.
* Updated _languagesMatch to compare candidate streams using the extracted resolved languages, ignoring non-descriptive 'und' matching.

### User Interface Components
* Explicit Flags Injection: Modified the calls to manager.playItems() and manager.setPendingItemOverrides() across all media types (Series, Season, Episode, BoxSet, defaultCase) to pass audioSelectionExplicit: viewModel.selectedAudioIndex != null and subtitleSelectionExplicit: viewModel.selectedSubtitleIndex != null.
* Initialization & Version Selection Fix: Corrected compilation errors where widget.selectedMediaSourceId was incorrectly accessed inside _ItemDetailScreenState. Removed widget.selectedMediaSourceId from _ItemDetailScreenState.initState() and _onChanged(), allowing the state variable _selectedMediaSourceId to properly initialize to the first available media source when the item loads.
 
## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code

## Testing

- [X] Manual testing completed: Tested persistence across Android TV and Windows Desktop clients.

### Test Steps
1. Navigate to the Media Info/Details page of an item with multiple versions, audio, or subtitle tracks.
2. Select a specific version, audio track, or subtitle track (e.g., Arabic subtitles when English are set as default).
3. Start playback without manually toggling it inside the player. Verify that the correct tracks/versions are applied.
4. Play subsequent episodes in the queue and verify that custom selections carry over dynamically based on language matching (e.g. Arabic subtitles carry over to subsequent episodes correctly).

N/A

## Checklist

### DIFF
--- a/lib/ui/screens/detail/item_detail_screen.dart
+++ b/lib/ui/screens/detail/item_detail_screen.dart
@@ -219,6 +219,15 @@ class _ItemDetailScreenState extends State<ItemDetailScreen>
         _lastFocusedBackdropItemId = null;
         _backgroundService.setBackground(item, context: BlurContext.details);
         _backdropUrl = _backgroundService.currentUrl;
+
+        if (item.mediaSources.isNotEmpty) {
+          _selectedMediaSourceId = item.mediaSources.first['Id'] as String?;
+        } else {
+          _selectedMediaSourceId = null;
+        }
+
+        _viewModel.selectedAudioIndex = null;
+        _viewModel.selectedSubtitleIndex = null;
       }
       if (!_themeMusicStarted) {
         _themeMusicStarted = true;
@@ -5983,6 +5992,8 @@ class _ActionButtonsState extends State<_ActionButtons> {
               startPosition: startPosition,
               audioStreamIndex: audioStreamIndex,
               subtitleStreamIndex: subtitleStreamIndex,
+              audioSelectionExplicit: viewModel.selectedAudioIndex != null,
+              subtitleSelectionExplicit: viewModel.selectedSubtitleIndex != null,
               enableDirectPlay: directAllowed,
               enableDirectStream: directAllowed,
             );
@@ -6015,6 +6026,8 @@ class _ActionButtonsState extends State<_ActionButtons> {
               startPosition: startPosition,
               audioStreamIndex: audioStreamIndex,
               subtitleStreamIndex: subtitleStreamIndex,
+              audioSelectionExplicit: viewModel.selectedAudioIndex != null,
+              subtitleSelectionExplicit: viewModel.selectedSubtitleIndex != null,
               enableDirectPlay: directAllowed,
               enableDirectStream: directAllowed,
             );
@@ -6070,6 +6083,8 @@ class _ActionButtonsState extends State<_ActionButtons> {
                 startPosition: startPosition,
                 audioStreamIndex: audioStreamIndex,
                 subtitleStreamIndex: subtitleStreamIndex,
+                audioSelectionExplicit: viewModel.selectedAudioIndex != null,
+                subtitleSelectionExplicit: viewModel.selectedSubtitleIndex != null,
                 mediaSourceId: widget.selectedMediaSourceId,
                 enableDirectPlay: directAllowed,
                 enableDirectStream: directAllowed,
@@ -6132,6 +6147,8 @@ class _ActionButtonsState extends State<_ActionButtons> {
               startPosition: startPosition,
               audioStreamIndex: audioStreamIndex,
               subtitleStreamIndex: subtitleStreamIndex,
+              audioSelectionExplicit: viewModel.selectedAudioIndex != null,
+              subtitleSelectionExplicit: viewModel.selectedSubtitleIndex != null,
               enableDirectPlay: directAllowed,
               enableDirectStream: directAllowed,
             );
@@ -6176,6 +6193,10 @@ class _ActionButtonsState extends State<_ActionButtons> {
               subtitleStreamIndex: applyMainItemStreamOverrides
                   ? subtitleStreamIndex
                   : null,
+              audioSelectionExplicit: applyMainItemStreamOverrides &&
+                  viewModel.selectedAudioIndex != null,
+              subtitleSelectionExplicit: applyMainItemStreamOverrides &&
+                  viewModel.selectedSubtitleIndex != null,
               mediaSourceId: applyMainItemStreamOverrides
                   ? selectedMediaSourceId
                   : null,
@@ -6188,6 +6209,8 @@ class _ActionButtonsState extends State<_ActionButtons> {
                 audioStreamIndex: audioStreamIndex,
                 subtitleStreamIndex: subtitleStreamIndex,
                 mediaSourceId: selectedMediaSourceId,
+                audioSelectionExplicit: viewModel.selectedAudioIndex != null,
+                subtitleSelectionExplicit: viewModel.selectedSubtitleIndex != null,
               );
             }
             await playItemsFuture;
--- a/lib/ui/screens/detail/item_detail_screen.dart
+++ b/lib/ui/screens/detail/item_detail_screen.dart
@@ -221,7 +221,7 @@ class _ItemDetailScreenState extends State<ItemDetailScreen>
         _backdropUrl = _backgroundService.currentUrl;
 
         if (item.mediaSources.isNotEmpty) {
-          _selectedMediaSourceId = item.mediaSources.first['Id'] as String?;
+          _selectedMediaSourceId = item.mediaSources.first['Id']?.toString();
         } else {
           _selectedMediaSourceId = null;
         }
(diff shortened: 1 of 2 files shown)


packages/playback_core/lib/src/playback_manager.dart
```diff
@@ -2279,3 +2371,284 @@ class PlaybackBringupState {
...
+  final normalized = language.trim().toLowerCase();
+  if (normalized.isEmpty) return '';
+  final base = normalized.split(RegExp(r'[-_]')).first;
+  if (base == 'arabic') return 'ara';
+  if (base == 'english') return 'eng';
+  return base;
+}
+
+String _toIso3(String language) {
+  if (language.length == 3) return language;
```
RadicalMuffinMan, line 2482:
```suggestion
  if (_kLanguageKeywords.containsKey(language)) return _kLanguageKeywords[language]!;
  if (language.length == 3) return language;
```

You already have a table (_kLanguageKeywords) that knows fre to fra, ger to deu, english to eng, etc., but nothing calls it during comparison. it's only used when reading a title. Routing _toIso3 through that table first means the two common 3-letter standards (B and T) and full words all collapse to one canonical code, so "French" picked on episode 1 still matches "fra" on episode 2.

packages/playback_core/lib/src/playback_manager.dart
```diff
@@ -2279,3 +2371,284 @@ class PlaybackBringupState {
...
+  'hindi': 'hin',
+  'hin': 'hin',
+};
+
+String _normalizeLanguage(String? language) {
+  if (language == null) return '';
+  final normalized = language.trim().toLowerCase();
+  if (normalized.isEmpty) return '';
+  final base = normalized.split(RegExp(r'[-_]')).first;
+  if (base == 'arabic') return 'ara';
```
RadicalMuffinMan:
```suggestion

```
not needed once the fix from line 2465 is implemented

packages/playback_core/lib/src/playback_manager.dart
```diff
@@ -2279,3 +2371,284 @@ class PlaybackBringupState {
...
+  if (base == 'english') return 'eng';
```
RadicalMuffinMan:
```suggestion

```
not needed once the fix from line 2465 is implemented

packages/playback_core/lib/src/playback_manager.dart
```diff
@@ -2279,3 +2371,284 @@ class PlaybackBringupState {
...
+  'heb': 'heb',
+  'hindi': 'hin',
+  'hin': 'hin',
+};
+
+String _normalizeLanguage(String? language) {
+  if (language == null) return '';
+  final normalized = language.trim().toLowerCase();
+  if (normalized.isEmpty) return '';
+  final base = normalized.split(RegExp(r'[-_]')).first;
```
RadicalMuffinMan:
```suggestion
  return normalized.split(RegExp(r'[-_]')).first;
```

Two functions (_normalizeLanguage and _extractLanguage) both try to translate language names, but only one of them is complete. Keep the translation in exactly one place (the keyword table) so there's no "why does Arabic work but Portuguese doesn't?" surprise later.

packages/playback_core/lib/src/playback_manager.dart
```diff
@@ -972,12 +983,92 @@ class PlaybackManager implements AudioOwnable {
...
+        if (_subtitleStreamIndex != matchedIdx) {
+          _subtitleStreamIndex = matchedIdx;
+          if (resolution.playMethod == StreamPlayMethod.transcode) {
+            needsReResolve = true;
+          }
+        }
+      }
+    }
+
+    if (needsReResolve) {
```
RadicalMuffinMan:
```suggestion
    iif (needsReResolve && !_reResolvingForTrackMatch) {
  _reResolvingForTrackMatch = true;
  try {
```

A function that re-calls itself should never depend on "the data will happen to be equal next time" to stop. A boolean flag guarantees it re-resolves at most once per play, so a weird server response can't turn into an infinite loop / reload storm. The try/finally makes sure the flag resets even if the re-resolve throws.

packages/playback_core/lib/src/playback_manager.dart
```diff
@@ -47,7 +47,10 @@ class PlaybackManager implements AudioOwnable {
   StreamResolutionResult? _lastPlaybackResolution;
```
RadicalMuffinMan, line 47:
```suggestion
  StreamResolutionResult? _lastPlaybackResolution;
  bool _reResolvingForTrackMatch = false;
```
for the fix at line 1027

packages/playback_core/lib/src/playback_manager.dart
```diff
@@ -972,12 +983,92 @@ class PlaybackManager implements AudioOwnable {
       return;
     }
 
+    bool needsReResolve = false;
+
+    if (_lastExplicitAudioLanguage != null) {
```
RadicalMuffinMan, line 993: The new audio/subtitle matching block is a near copy of the block this same PR edits at lines 2140/2154. Extract one helper and call it from both:
```
int? _matchStreamIndexByLanguage(List<Map<String, dynamic>> streams, String? lang, String type) {
  final candidates = streams.where((s) => s['Type'] == type).toList();
  final i = candidates.indexWhere((s) => _languagesMatch(s, lang));
  return i >= 0 ? candidates[i]['Index'] as int? : null;
}
```

packages/playback_core/lib/src/playback_manager.dart
```diff
@@ -972,12 +983,92 @@ class PlaybackManager implements AudioOwnable {
...
+      final audioStreams = resolution.mediaStreams.where((s) => s['Type'] == 'Audio').toList();
+      final targetIdx = audioStreams.indexWhere(
+        (s) => _languagesMatch(s, _lastExplicitAudioLanguage),
+      );
+      if (targetIdx >= 0) {
+        final matchedIdx = audioStreams[targetIdx]['Index'] as int?;
+        if (_audioStreamIndex != matchedIdx) {
```
RadicalMuffinMan:
```suggestion
        if (matchedIdx != null && _audioStreamIndex != matchedIdx) {
```
"I found the right track" and "the track has no usable index" are different outcomes. it should only act when it actually has an index, otherwise you can replace a good selection with nothing.

lib/ui/screens/detail/item_detail_screen.dart
```diff
@@ -219,6 +219,15 @@ class _ItemDetailScreenState extends State<ItemDetailScreen>
         _lastFocusedBackdropItemId = null;
         _backgroundService.setBackground(item, context: BlurContext.details);
         _backdropUrl = _backgroundService.currentUrl;
+
+        if (item.mediaSources.isNotEmpty) {
+          _selectedMediaSourceId = item.mediaSources.first['Id'] as String?;
```
RadicalMuffinMan:
```suggestion
          _selectedMediaSourceId = item.mediaSources.first['Id']?.toString()
```
as String? throws if Id ever comes back as a non-string (like an int from a different server). item.mediaSources.first['Id']?.toString() can't throw and gives the same result. Cheap insurance on data you don't control.

RadicalMuffinMan (changes requested):

packages/playback_core/lib/src/playback_manager.dart
mattsigal: Helper committed!

PR #540, opened 2026-06-15
fix(playback): fix built-in speaker detection and bypass capabilities…

mattsigal:
# Pull Request

## Summary
This PR fixes the issue where built-in TV speakers are misdetected as HDMI receivers (ROUTE_HDMI), and when the user manually overrides "Audio Output" to "AV receiver (Atmos / DTS:X)", the app still downmixes to stereo. It updates the native audio routing logic to correctly detect TV speakers and shifts the capability-gating logic to the Dart side so that manual preference overrides can bypass route restrictions.

On a Smart TV panel (e.g., Sony Bravia, TCL, Hisense running Android TV) using its built-in speakers the Android OS reports both AudioDeviceInfo.TYPE_BUILTIN_SPEAKER and AudioDeviceInfo.TYPE_HDMI (the HDMI inputs on the back of the TV panel). Before this PR, because TYPE_HDMI was checked first in resolveRouteType, the TV resolved to ROUTE_HDMI (thinking an external AVR was connected). As a result, the TV speakers were incorrectly treated as a surround receiver, and the app attempted to bitstream lossless formats like TrueHD and DTS-HD. Because TV speakers cannot decode or handle lossless passthrough, the user got complete silence or playback hangs. 

With this PR, isExternalBox returns false on a TV panel, causing it to correctly fall back to ROUTE_SPEAKER (TV Speakers), restricting the output to stereo PCM and forcing the server to transcode TrueHD/DTS-HD.

## Related Issues
* TV built-in speakers detected as ROUTE_HDMI instead of ROUTE_SPEAKER (limiting channels to 2 and disabling TrueHD/DTS-HD).
* Manual override to "AV receiver (Atmos / DTS:X)" still downmixing to stereo or failing to pass through HD audio.
* Discord user issue using TV speakesr

- Closes #
- Fixes #
- Related to #

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
Android Native (android/app) AudioCapabilities.kt :
* Removed native route-based capability clearing block (TrueHD/DTS-HD/Atmos).
* Added isExternalBox() check using Build.BRAND, Build.MANUFACTURER, and Build.MODEL to skip identifying dummy speakers on external TV boxes (Shield, Chromecast, Fire TV, Mi Box, Onn).
* Updated resolveRouteType to prioritize bluetooth, wired/USB headsets, ARC/eARC, built-in speakers (on non-external panels), HDMI/LINE_DIGITAL, and then built-in speakers as a fallback.

Flutter/Dart Client (lib) audio_capability_profile.dart:
* Shifted ARC/speaker capability gating logic from native code to Dart side in AudioCapabilityProfile.fromMap.
* Accept AudioOutputMode in fromMap. If AudioOutputMode.avrPassthrough is active, override the route type to AudioRouteType.hdmi (if not already earc), and force maxPcmChannels to at least 8, bypassing speaker route restrictions and preserving detected receiver pass-through capabilities.

user_preferences.dart:
* Updated detectedAudioCapabilities to resolve and pass the current AudioOutputMode to AudioCapabilityProfile.fromMap.

settings_side_panel.dart:
* Updated settings screen's _audioCapabilityProfile helper to pass the resolved AudioOutputMode.

Player Backends & Playback Module:
Refactored  media3_player_backend.dart, media_kit_player_backend.dart, appletv_mpv_backend.dart, html_video_backend_profile.dart, and playback_module.dart to query the consolidated detectedAudioCapabilities from preferences, removing redundant snapshot parses and unused imports.

## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code

## Testing

### Test Steps
1. Deploy to Shield TV. On "Auto", verify routing resolves to ROUTE_HDMI (showing "HDMI • 8ch PCM" or passthrough formats).
2. Deploy to a TV with built-in speakers. On "Auto", verify routing resolves to ROUTE_SPEAKER (limiting PCM channels to 2 and hiding TrueHD/DTS-HD options).
3. Toggle "Audio Output" to "AV receiver (Atmos / DTS:X)" on the TV. Verify capabilities override to HDMI / 8 PCM channels and expose passthrough capabilities.

AVR  ON:
Shield_Screenshot_2026-06-14_22-33-50

AVR OFF:
Shield_Screenshot_2026-06-14_22-35-36

## Checklist

### DIFF
--- a/android/app/src/main/kotlin/org/moonfin/androidtv/AudioCapabilities.kt
+++ b/android/app/src/main/kotlin/org/moonfin/androidtv/AudioCapabilities.kt
@@ -227,7 +227,7 @@ object AudioCapabilities {
         val supportsDts = canPassthroughDts || canPassthroughDtsHd || canPassthroughDtsX
         val supportsTrueHd = canPassthroughTrueHd || canPassthroughTrueHdJoc
 
-        val maxPcmChannels = estimateMaxPcmChannels(routeType)
+        val maxPcmChannels = detectMaxPcmChannels(bitstreamDevices, routeType)
 
         return mapOf(
             "supportsAc3" to supportsAc3,
@@ -418,51 +418,44 @@ object AudioCapabilities {
             return ROUTE_ARC
         }
 
-        // 4. Built-in speakers. Only route to Speaker if the device is a Smart TV panel
-        // (not an external set-top box/dongle like Shield or Chromecast, which advertise
-        // dummy built-in speakers).
-        if (!isExternalBox()) {
-            if (types.contains(AudioDeviceInfo.TYPE_BUILTIN_SPEAKER) ||
-                types.contains(AudioDeviceInfo.TYPE_BUILTIN_EARPIECE)
-            ) {
-                return ROUTE_SPEAKER
-            }
-        }
-
-        // 5. Standard HDMI (e.g. for external streaming boxes like Nvidia Shield,
-        // Chromecast, Fire Stick, which have no built-in speakers).
-        if (types.contains(AudioDeviceInfo.TYPE_HDMI) ||
-            types.contains(AudioDeviceInfo.TYPE_LINE_DIGITAL)
-        ) {
-            return ROUTE_HDMI
-        }
-
-        // Fallback for TV panels if we didn't match HDMI above but TYPE_BUILTIN_SPEAKER
-        // was skipped.
+        // 4. Built-in speakers default to a safe stereo route. We prefer this
+        // over plain HDMI on purpose, so a TV panel never tries to bitstream
+        // lossless audio to speakers that cannot decode it, which is the silent
+        // playback bug. A box wired to a real AV receiver should turn on the
+        // "AV receiver" output mode, which forces the HDMI route and passthrough
+        // on the Dart side.
         if (types.contains(AudioDeviceInfo.TYPE_BUILTIN_SPEAKER) ||
             types.contains(AudioDeviceInfo.TYPE_BUILTIN_EARPIECE)
         ) {
             return ROUTE_SPEAKER
         }
 
+        // 5. Standard HDMI for boxes with no built-in speaker.
+        if (types.contains(AudioDeviceInfo.TYPE_HDMI) ||
+            types.contains(AudioDeviceInfo.TYPE_LINE_DIGITAL)
+        ) {
+            return ROUTE_HDMI
+        }
+
         return ROUTE_OTHER
     }
 
-    private fun isExternalBox(): Boolean {
-        val brand = Build.BRAND.lowercase()
-        val manufacturer = Build.MANUFACTURER.lowercase()
-        val model = Build.MODEL.lowercase()
-        
-        return brand.contains("nvidia") || 
-               brand.contains("google") || 
-               manufacturer.contains("nvidia") ||
-               manufacturer.contains("google") ||
-               model.contains("shield") || 
-               model.contains("chromecast") || 
-               model.contains("mi box") || 
-               model.contains("mibox") ||
-               model.contains("fire tv") ||
-               model.contains("onn")
+    private fun detectMaxPcmChannels(devices: List<AudioDeviceInfo>, routeType: String): Int {
+        var maxVal = 0
+        for (device in devices) {
+            val counts = device.channelCounts
+            if (counts != null) {
+                for (count in counts) {
+                    if (count > maxVal) {
+                        maxVal = count
+                    }
+                }
+            }
+        }
+        if (maxVal > 0) {
+            return maxVal
+        }
+        return estimateMaxPcmChannels(routeType)
     }
 
     private fun estimateMaxPcmChannels(routeType: String): Int {
--- a/android/app/src/main/kotlin/org/moonfin/androidtv/AudioCapabilities.kt
+++ b/android/app/src/main/kotlin/org/moonfin/androidtv/AudioCapabilities.kt
@@ -223,7 +223,6 @@ object AudioCapabilities {
         // turn the TrueHD-Atmos passthrough toggle on explicitly (override wins).
 
 
-
         val supportsAc3 = canPassthroughAc3 || canPassthroughEac3
         val supportsDts = canPassthroughDts || canPassthroughDtsHd || canPassthroughDtsX
         val supportsTrueHd = canPassthroughTrueHd || canPassthroughTrueHdJoc
--- a/android/app/src/main/kotlin/org/moonfin/androidtv/AudioCapabilities.kt
+++ b/android/app/src/main/kotlin/org/moonfin/androidtv/AudioCapabilities.kt
@@ -222,17 +222,7 @@ object AudioCapabilities {
         // AVR can't handle. Users who know their receiver supports it can still
         // turn the TrueHD-Atmos passthrough toggle on explicitly (override wins).
 
-        // ARC (plain Audio Return Channel) only carries compressed audio. Force
-        // the lossless/HD caps off so we never advertise TrueHD / TrueHD-Atmos /
-        // DTS-HD / DTS:X over a plain ARC link. eARC and direct HDMI keep them.
-        // (Mirrored as a chokepoint in AudioCapabilityProfile.fromMap on the
-        // Dart side.)
-        if (routeType == ROUTE_ARC) {
-            canPassthroughTrueHd = false
-            canPassthroughTrueHdJoc = false
-            canPassthroughDtsHd = false
-            canPassthroughDtsX = false
-        }
+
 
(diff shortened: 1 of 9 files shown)


lib/preference/user_preferences.dart
```diff
@@ -306,6 +211,7 @@ class UserPreferences extends ChangeNotifier {
       PlatformDetection.hasAudioCapabilities
       ? AudioCapabilityProfile.fromMap(
           PlatformDetection.audioCapabilitiesSnapshot,
+          audioOutputMode: resolveAudioOutputMode(),
         )
       : const AudioCapabilityProfile.optimistic();
```
RadicalMuffinMan: Change lines 210 - 216 to
```
  AudioCapabilityProfile get detectedAudioCapabilities =>
      PlatformDetection.hasAudioCapabilities
      ? AudioCapabilityProfile.fromMap(
          PlatformDetection.audioCapabilitiesSnapshot,
          audioOutputMode: resolveAudioOutputMode(),
        )
      : PlatformDetection.isIOS
          ? const AudioCapabilityProfile.appleMobile()
          : const AudioCapabilityProfile.optimistic();
```
This getter is now the only place that builds the profile, but it only falls back to optimistic(). media_kit used to fall back to appleMobile() on iOS, and that was the single user of appleMobile(). iOS has no native audio probe, so hasAudioCapabilities is false there, which means after this change iOS gets optimistic(), and optimistic() sets canDecodeTrueHd to true while appleMobile() sets it to false. That makes iOS claim it can decode TrueHD when it cannot, so the server is never asked to transcode it and you get silent or failed TrueHD on iPhone and iPad. Keeping the iOS branch here preserves the old behavior.

android/app/src/main/kotlin/org/moonfin/androidtv/AudioCapabilities.kt
```diff
@@ -222,17 +222,7 @@ object AudioCapabilities {
...
-        // DTS-HD / DTS:X over a plain ARC link. eARC and direct HDMI keep them.
-        // (Mirrored as a chokepoint in AudioCapabilityProfile.fromMap on the
-        // Dart side.)
-        if (routeType == ROUTE_ARC) {
-            canPassthroughTrueHd = false
-            canPassthroughTrueHdJoc = false
-            canPassthroughDtsHd = false
-            canPassthroughDtsX = false
-        }
+
```
RadicalMuffinMan, line 225:
```suggestion
```

android/app/src/main/kotlin/org/moonfin/androidtv/AudioCapabilities.kt
```diff
@@ -414,21 +418,54 @@ object AudioCapabilities {
...
-        }
-        if (types.any(::isSpeakerLikeType)) {
+
+        // Fallback for TV panels if we didn't match HDMI above but TYPE_BUILTIN_SPEAKER
+        // was skipped.
+        if (types.contains(AudioDeviceInfo.TYPE_BUILTIN_SPEAKER) ||
+            types.contains(AudioDeviceInfo.TYPE_BUILTIN_EARPIECE)
+        ) {
             return ROUTE_SPEAKER
         }
```
RadicalMuffinMan: change 422 - 447 of this to 

```
        // 4. Built-in speakers default to a safe stereo route. We prefer this
        // over plain HDMI on purpose, so a TV panel never tries to bitstream
        // lossless audio to speakers that cannot decode it, which is the silent
        // playback bug. A box wired to a real AV receiver should turn on the
        // "AV receiver" output mode, which forces the HDMI route and passthrough
        // on the Dart side.
        if (types.contains(AudioDeviceInfo.TYPE_BUILTIN_SPEAKER) ||
            types.contains(AudioDeviceInfo.TYPE_BUILTIN_EARPIECE)
        ) {
            return ROUTE_SPEAKER
        }

        // 5. Standard HDMI for boxes with no built-in speaker.
        if (types.contains(AudioDeviceInfo.TYPE_HDMI) ||
            types.contains(AudioDeviceInfo.TYPE_LINE_DIGITAL)
        ) {
            return ROUTE_HDMI
        }
```

This drops the per manufacturer isExternalBox() check. The reason it existed makes sense, a TV panel and an external box both report a built-in speaker plus HDMI, and you want opposite routing, so device types alone cannot tell them apart, which is why the you tried doing a brand list. The problem is the list is an allowlist, so any box that is not on it (Beelink, Ugoos, Formuler, Homatics, generic boxes, the new Google TV Streamer) falls through to speaker anyway and loses surround on a receiver, and it needs editing for every new device. Instead, default a built-in speaker to stereo, which is always audible and never silent, and let the "AV receiver" output mode this PR already adds be the opt-in for surround. The tradeoff is that a box wired to an AVR now needs that setting flipped once, but there is no device census to maintain and it never trades silence for surround in either direction. The old fallback speaker check below is now redundant, so it is folded in here.

android/app/src/main/kotlin/org/moonfin/androidtv/AudioCapabilities.kt
```diff
@@ -414,21 +418,54 @@ object AudioCapabilities {
...
+               brand.contains("google") || 
+               manufacturer.contains("nvidia") ||
+               manufacturer.contains("google") ||
+               model.contains("shield") || 
+               model.contains("chromecast") || 
+               model.contains("mi box") || 
+               model.contains("mibox") ||
+               model.contains("fire tv") ||
+               model.contains("onn")
+    }
```
RadicalMuffinMan, line 459:
```suggestion
```
With the route logic above no longer calling it, isExternalBox() is dead code, so remove it.

lib/playback/audio_capability_profile.dart
```diff
@@ -158,18 +171,23 @@ class AudioCapabilityProfile {
...
-          !isArc &&
+          isHdRoute &&
           _readBool(values, 'canPassthroughTrueHdJoc', defaultValue: false),
-      maxPcmChannels: _readInt(values, 'maxPcmChannels', defaultValue: 8),
+      maxPcmChannels: (() {
+        final rawMaxChannels = _readInt(values, 'maxPcmChannels', defaultValue: 8);
+        return audioOutputMode == AudioOutputMode.avrPassthrough && rawMaxChannels < 8
+            ? 8
+            : rawMaxChannels;
+      })(),
```
RadicalMuffinMan:
```suggestion
      maxPcmChannels:
          audioOutputMode == AudioOutputMode.avrPassthrough &&
              _readInt(values, 'maxPcmChannels', defaultValue: 8) < 8
          ? 8
          : _readInt(values, 'maxPcmChannels', defaultValue: 8),
```

this just swaps the immediately invoked closure for a plain ternary, which reads a little easier. It reads the map twice but that is a trivial lookup. Keep the avrPassthrough behavior as is. The isHdRoute change just above this is good, leave it.

lib/playback/appletv_mpv_backend.dart
```diff
@@ -6,7 +6,7 @@ import 'package:playback_core/playback_core.dart';
 import '../preference/preference_constants.dart';
 import '../preference/user_preferences.dart';
 import '../util/platform_detection.dart';
-import 'audio_capability_profile.dart';
+
```
RadicalMuffinMan, line 10:
```suggestion
import 'audio_capability_profile.dart';
```
This import was removed, but the file still uses AudioRouteType at line 336 (and AudioCapabilityProfile at 329 and 332), so the Apple TV target no longer compiles. Even after the next change, AudioRouteType is still referenced here, so the import has to come back.

lib/playback/appletv_mpv_backend.dart
RadicalMuffinMan:
```suggestion
    final profile = _prefs.detectedAudioCapabilities;
```

This is the second profile site in the file and the refactor missed it. Routing it through the same getter as everything else fixes the compile reference to AudioCapabilityProfile and, more importantly, makes the AV receiver override actually apply here. Right now this path reads the raw snapshot, so a user who overrides to AV receiver passthrough would still get stereo channel selection on Apple TV because maxPcmChannels was never bumped. The getter applies that override, so after this the channel mode follows it. The forceStereo early return just above still wins first, which is correct.

RadicalMuffinMan (changes requested):
Rebase to latest and fix merge conflicts then implement the fixes below. this renders most of #642 unnecessary

mattsigal:
All comments have been addressed (and elements of #642 included, so that PR has been closed).

RadicalMuffinMan:
Apple tv and ios fixes aren't added and the isExternalBox() is still there, also all builds are failing

PR #557, opened 2026-06-16
Defeat the jiggle during Expanded Home Row Navigation

mattsigal:
# Pull Request

## Summary
Resolves the home row vertical scrolling "jiggle" when navigating between rows or snapping on idle. The jiggle was caused by target layouts being calculated dynamically from live render heights that changed during focus shift animation (`AnimatedPadding`). Shifting layout and scroll offset calculations to use a static row height helper eliminates the layout shifts and ensures smooth, stable transitions.

## Related Issues
Too jiggly.

- Closes #
- Fixes #
- Related to #

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [X] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
- Implemented `_staticRowHeight(int rowIndex)` in `home_screen.dart` to calculate the exact row widget heights statically based on preferences and row types.
- Refactored `_tvTargetTopForRow` and `_estimatedRowExtent` to query the static row height helper instead of the dynamic render tree.
- Simplified `_scrollTvRowIntoOverlayBand` to animate directly to pre-calculated offsets (`_rowTopOffsets[rowIndex] - targetTop`), removing the reliance on dynamic global coordinates.
- Disabled `focusedRowSpacing` (setting it to `0.0`) and the vertical `AnimatedPadding` vertical padding/spacing on the focused row when `fullScreenRows` is active, since adjacent rows are hidden.

## Platform
- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code

## Testing

### Test Steps
1. Configure home screen to use Modern TV style and enable "Expanded Home Rows" (fullScreenRows).
2. Navigate vertically using a remote (D-pad) or keyboard. Verify that transition between rows is smooth, immediate, and does not wobble or overshoot.
3. Scroll with a mouse wheel and verify that snapping pulls the layout cleanly to the target row without settle adjustments.

## Checklist

### DIFF
--- a/lib/ui/screens/home/home_screen.dart
+++ b/lib/ui/screens/home/home_screen.dart
@@ -1900,14 +1900,69 @@ class _ContentRowsState extends State<_ContentRows>
     return _rowKeys[rowIndex]?.currentState as LockedFocusRowState?;
   }
 
+  double _staticRowHeight(int rowIndex) {
+    final row = rowIndex < widget.viewModel.rows.length ? widget.viewModel.rows[rowIndex] : null;
+    if (row == null) return 0.0;
+
+    final prefs = widget.prefs;
+    final posterSize =
+        (_isHomeRowsStyleV2() &&
+            !prefs.containsPreference(UserPreferences.posterSize))
+        ? PosterSize.small
+        : prefs.get(UserPreferences.posterSize);
+
+    final desktopScale = _desktopUiScaleFactor();
+    final metadataScale = desktopScale;
+    final isRowsV2 = prefs.get(UserPreferences.homeRowsStyle) == HomeRowsStyle.v2 && !_isSeerrFilterRow(row);
+
+    double childHeight = 0.0;
+    if (row.isLoading) {
+      childHeight = 220.0 * metadataScale;
+    } else if (row.rowType == HomeRowType.liveTv ||
+        row.rowType == HomeRowType.libraryTilesSmall) {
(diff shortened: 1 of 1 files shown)


RadicalMuffinMan:
please fix the merge conflicts

lib/ui/screens/home/home_screen.dart
```diff
@@ -1969,31 +1993,13 @@ class _ContentRowsState extends State<_ContentRows>
   Future<void> _scrollTvRowIntoOverlayBand(int rowIndex) async {
     if (!mounted || !_scrollController.hasClients) return;
```
RadicalMuffinMan, line 2014:
```suggestion
    if (!mounted || !_scrollController.hasClients) return;
     if (rowIndex < 0 || rowIndex >= _rowTopOffsets.length) return;
```

The old code measured the render tree and returned early when the container context was null, which implicitly covered "not laid out yet." This version reads _rowTopOffsets[rowIndex] directly with no such guard, so any moment where focus lands on a row that isn't in the offsets list yet is a hard crash. The guard restores the old early-return behavior.

RadicalMuffinMan (changes requested):
there is also still a merge conflict

PR #604, opened 2026-06-21
UI adjustments & video_player new features

chwps:
# Pull Request

## Summary
This PR introduces several quality-of-life improvements and fixes to the Video Player. Updates include a smarter "Next Up" logic that protects post-credits scenes, auto-hiding behavior for skip widgets, and UI adjustments to ensure transport controls remain perfectly centered across all states. It also addresses a controls overlay bug specific to the web platform.

## Type of Change
- [X] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Performance improvement
- [X] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

**UI Adjustments:**
- **Centered Transport Controls:** Introduced a `hasAnyNavigation` logic. The previous/next queue arrows are now always rendered in the layout (grayed out when unavailable). This prevents layout shifts and ensures the center transport controls (Play/Pause, Seek forward/backward) remain perfectly centered on screen.
- **Overlay Positioning:** Refined the vertical alignment of the Skip and Next Up widgets. Previously too high on mobile and too low on desktop, now optimized per platform ratio (Desktop: 140.0, Mobile: 100.0, TV: 190.0). (Not tested on every devices)

**New Features:**
- **Auto-Hide Prompts:** "Skip Intro/Outro" and "Next Up" widgets now automatically hide after 10 seconds of playback. They reappear if the user brings up the controls overlay.
- **Post-Credits Scene Protection:** The "Replace Skip Outro with Next Up" feature now checks for remaining video duration. If content exists after the outro segment, it falls back to the standard "Skip Outro" button to prevent users from missing post-credits scenes. 
- **New Setting:** Added a dependent setting ("Next Up Ignore Remaining Content") to allow users to bypass the post-credits protection and force the Next Up overlay if preferred.

**Bug Fixes:**
- **Web Interface:** Fixed an issue where clicking directly on the video player failed to dismiss the controls overlay.

## Platform
- [ ] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code

## Testing

### Test Steps
1. Navigate through media with and without previous/next queue items to verify transport control centering.
2. Trigger outro segments with and without post-credits content to test the new Next Up protection logic.
3. Verify the new dependent setting toggles correctly in the settings panel.
4. Let the Skip/Next Up overlays time out (10s) and verify they reappear with the controls overlay.
5. (Web) Click on the video surface to ensure the controls overlay dismisses properly.

## Checklist

### DIFF
--- a/devtools_options.yaml
+++ b/devtools_options.yaml
@@ -1,0 +1,3 @@
+description: This file stores settings for Dart & Flutter DevTools.
+documentation: https://docs.flutter.dev/tools/devtools/extensions#configure-extension-enablement-states
+extensions:
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -7230,7 +7230,7 @@
   "alwaysOn": "Always on.",
   "replaceSkipOutroWithNextUpDisplay": "Replace Skip Outro with Next Up Display",
   "replaceSkipOutroWithNextUpDisplaySubtitle": "Show the Next Up overlay instead of the Skip Outro button.",
-  "nextUpIgnoreRemainingContentDisplay": "Play Next (Ignore Remaining Content After Outro)",
+  "nextUpIgnoreRemainingContentDisplay": "Next Up Forced (Ignore Remaining Content After Outro)",
   "nextUpIgnoreRemainingContentDisplaySubtitle": "Trigger Next Up pop-up even if there's remaining content after outro.",
   "playerRouting": "Player Routing",
   "preferSoftwareDecoders": "Prefer software decoders",
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -7230,6 +7230,8 @@
   "alwaysOn": "Always on.",
   "replaceSkipOutroWithNextUpDisplay": "Replace Skip Outro with Next Up Display",
   "replaceSkipOutroWithNextUpDisplaySubtitle": "Show the Next Up overlay instead of the Skip Outro button.",
+  "nextUpIgnoreRemainingContentDisplay": "Play Next (Ignore Remaining Content After Outro)",
+  "nextUpIgnoreRemainingContentDisplaySubtitle": "Trigger Next Up pop-up even if there's remaining content after outro.",
   "playerRouting": "Player Routing",
   "preferSoftwareDecoders": "Prefer software decoders",
   "preferSoftwareDecodersSubtitle": "Use FFmpeg (audio) and libgav1 (AV1) before hardware decoders. Disable if HDMI audio passthrough breaks.",
--- a/lib/l10n/app_localizations.dart
+++ b/lib/l10n/app_localizations.dart
@@ -11494,6 +11494,18 @@ abstract class AppLocalizations {
   /// **'No Jellyfin servers reporting the plugin yet.'**
   String get homeScreenSectionsIntegrationNoServers;
 
+  /// No description provided for @kefinTweaksIntegrationDescription.
+  ///
+  /// In en, this message translates to:
+  /// **'Detect rows configured via ranaldsgift\'s \"KefinTweaks\" plugin. Custom sections, recently released, watch again, seasonal, and recently added in library are mirrored from the KefinTweaks configuration on each Jellyfin server.'**
+  String get kefinTweaksIntegrationDescription;
+
+  /// No description provided for @kefinTweaksIntegrationNoServers.
+  ///
+  /// In en, this message translates to:
+  /// **'No Jellyfin servers reporting KefinTweaks yet.'**
+  String get kefinTweaksIntegrationNoServers;
+
   /// No description provided for @integrationOpenHomeSections.
   ///
   /// In en, this message translates to:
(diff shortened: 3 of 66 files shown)


mattsigal:
"Auto-Hide Prompts: "Skip Intro/Outro" and "Next Up" widgets now automatically hide after 10 seconds of playback. They reappear if the user brings up the controls overlay." -- There is already a setting for Next Up: "Next Up Timeout" which should be respected beyond the 10 seconds.

chwps:
Even if it's hidden from the player screen, the countdown will still continue, it's just to not have the widget always on screen if you want to see the media playing during this time. But you're absolutely right for Next Up setting, for Skip Intro/Outro however there were no such settings.

mattsigal:
Yes, that is what I was getting at - for the opening/closing credits, the hide/show behaviour is probably good (e.g., a 2-minute long opening credits sequence, might want to watch part of it then skip); but for the Next Up display which triggers after playback has ended, it shouldn't disappear since it is the only thing active on the screen.

chwps:
You are right I totally forgot about that. What do you think about this fix? It checks if it's at the end of the media with a margin of 2sec and if it is it doesn't launch the timer:

Old code:
<pre><code>
  // Launch auto-hide timer (10s) for SkipSegment.
  void _startSkipSegmentAutoHide() {
    _skipSegmentHideTimer?.cancel();
    if (_skipSegment == null) return;
    _skipSegmentHideTimer = Timer(const Duration(seconds: 10), () {
      if (mounted) {
        setState(() => _skipSegmentAutoHidden = true);
      }
    });
  }

  // Launch auto-hide timer (10s) for NextUp.
  void _startNextUpAutoHide() {
    _nextUpHideTimer?.cancel();
    if (_showNextUp == false || _nextUpItem == null) return;
    _nextUpHideTimer = Timer(const Duration(seconds: 10), () {
      if (mounted) {
        setState(() => _nextUpAutoHidden = true);
      }
    });
  }
</code></pre>
New code:
<pre><code>
  // Launch auto-hide timer (10s) for NextUp.
  void _startNextUpAutoHide() {
    _nextUpHideTimer?.cancel();
    if (_showNextUp == false || _nextUpItem == null) return;

    final duration = _state.duration;
    final position = _state.position;

    if (duration > Duration.zero) {
      final remaining = duration - position;
      if (remaining <= const Duration(seconds: 2)) {
        return; 
      }
    }
    _nextUpHideTimer = Timer(const Duration(seconds: 10), () {
      if (mounted) {
        setState(() => _nextUpAutoHidden = true);
      }
    });
  }
</code></pre>
Tried it and it works, I also corrected the comments that were in French sorry.

mattsigal:
I'm just providing some input but it is up to @RadicalMuffinMan to review the PR and see if he agrees. 

I still don't think your new logic works perfectly though -- since many users use the "Replace Skip Outro with Next Up Display" toggle which will trigger outside of the interval you've indicated. The auto-hide should simply never apply to the Next Up display, regardless of when it triggers.

chwps:
Ok I see! My logic was just to hide the big widget that was in the middle of the screen (especially on mobile) just so you can watch the outro in the meantime (I know you can use the minimal one that's smaller).
Since it will always trigger the skip when the countdown finishes, even if it's hidden, and since it reappears with the controls overlay anyway. 

But I totally understand your way of seeing it. 
Thank you for your insight, I really appreciate it, let's see what @RadicalMuffinMan says!

mattsigal:
If you don't want to see the big widget, you should leave the "Replace Skip Outro with Next Up Display" setting off. Then during the outro, you would only get a small "Skip Outro?" popup. That's the whole point of that setting.

chwps:
I think some users, like myself, still like to have the countdown of the next_up widget so you don't have to interact with the remote or the screen to first skip_outro and then next_up. But it's a minor inconvenience really, you can call that laziness :) 
And the more I think of it, I realize my countdown was less than 10sec anyway, so it would not have the time to hide.

I started to develop it to hide the skip intro/outro button after 10sec and then added it to Next Up to have the same behavior for everything.

Let's see what RadicalMuffinMan says, but I will happily remove this behavior if you two prefer!

RadicalMuffinMan:
Hey thanks for the PR and for the really constructive back and forth with @mattsigal along the way. Sorry it sat open so long.

Most of this has been overtaken since June. The auto-hide for the Skip Intro/Outro button landed on main as a setting (5s, 10s, or off, defaulting to 5s). The center transport controls no longer drift when there's no previous or next item, the row now balances them in equal flex slots so play/pause stays pinned in the center and for Next Up we ended up agreeing with the point raised here that it shouldn't auto-hide, since it's the only thing on screen once playback ends.

The player screen and the settings panel have both been reworked fairly heavily since then, so this branch is well past a clean rebase and I'm going to close it here rather than leave it hanging.

The one thing still worth chasing is the click-to-dismiss behavior on web if you'd like to submit a small standalone PR for just that, it would be very welcome.

Thanks again for digging into this, and sorry for the outcome after the work you put in.

chwps:
Hi,
No problem I saw that most of this was nicely implemented since!
Yes, I can definitely submit a PR for the click-to-dismiss on web, I will check that in the next few days.

If I can share my view for the auto hide of the skip button, I think it would be nice if it reappears with the control overlay after it diseappeared. Like: skip button appears -> auto hide after the set duration -> still visible and can still be interacted with in the controls overlay.
That way if you missed it, you can still click it afterwards.

If your interested in that behavior I can have a look and also submit a PR.

RadicalMuffinMan:
Yeah sounds great! Looking forward to it.

PR #628, opened 2026-06-23
Fixed pagination of Favorites Horizontal scroll

Brady3035:
# Pull Request

## Summary
Implements infinite scrolling pagination for favorites grouped into rows on the Home View. The previous approach relied on index tracking via item focus, which failed to page correctly on D-pad layout environments and constrained desktop/mobile interfaces, causing list rows to arbitrarily cap at the initial 30 items. 

By migrating to a pixel-exact scroll notification tracker, subsequent pages are automatically and reliably fetched from the Emby server as the user pans or navigates toward the end of a row.

## Related Issues
No open issues

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [x] Performance improvement
- [x] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
- Wrapped `LockedFocusRow` within a `NotificationListener<ScrollNotification>` inside the `_buildRows()` loop[cite: 1].
- Implemented axis metric validation to detect when horizontal scrolling gets within 500 pixels of the current row's right edge.
- Connected the scroll boundary trigger to `FavoritesViewModel.loadMoreForType(type)` to request the next data window asynchronously[cite: 2].
- Removed the old index-based pagination threshold check from the `onIndexChanged` callback to prevent duplicate trigger conflicts[cite: 1].

## Platform
- [ ] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [x] All / Shared code

## Testing

### Test Steps
1. Navigate to the Favorites Screen and ensure the view is configured to **Home View**.
2. Scroll horizontally or use D-pad navigation to move right across a populated row (e.g., Movies or Series).
3. Verify that once you approach the right margin, the UI seamlessly fetches and populates items beyond the initial 30-item constraint without losing focus state.

*N/A*

## Checklist

### DIFF
--- a/lib/ui/screens/browse/favorites_screen.dart
+++ b/lib/ui/screens/browse/favorites_screen.dart
@@ -364,82 +364,94 @@ class _FavoritesScreenState extends State<FavoritesScreen> with GridFocusNodeMix
             ),
             contentSpacing: 0,
             showControls: !isMobile && PlatformDetection.useDesktopUi,
-            builder: (context, scrollController) => LockedFocusRow<AggregatedItem>(
-              key: _rowKeys[type],
-              items: items,
-              hubKey: 'favorites_row_${type.name}',
-              controller: scrollController,
-              height: rowHeight,
-              itemExtent: imageH * ar,
-              itemSpacing: 12 * desktopScale,
-              padding: EdgeInsets.fromLTRB(
-                20 * desktopScale,
-                5 * desktopScale,
-                20 * desktopScale,
-                5 * desktopScale,
-              ),
-              onIndexChanged: (index, item) {
-                _onItemFocused(item);
-                if (index >= items.length - 8) {
-                  _vm.loadMoreForType(type);
+            builder: (context, scrollController) => NotificationListener<ScrollNotification>(
+              onNotification: (notification) {
+                // 1. Check if the scroll notification comes from the horizontal row
+                if (notification.metrics.axis == Axis.horizontal) {
+                  final maxScroll = notification.metrics.maxScrollExtent;
+                  final currentScroll = notification.metrics.pixels;
+                  
+                  // 2. Trigger a load when the user scrolls within 500 pixels of the edge
+                  if (maxScroll - currentScroll < 500) {
+                    _vm.loadMoreForType(type);
+                  }
                 }
+                return false; 
               },
-              onVerticalNavigation: (isUp) {
-                return _onRowVerticalNavigation(visibleTypes, type, isUp);
-              },
-              onLongPress: (index, item) => showContextMenu(
-                context,
-                item,
-                onChanged: () => setState(() {}),
-              ),
-              onTap: (index, item) => context.push(
-                Destinations.itemOrPhoto(
-                  item.id,
-                  serverId: item.serverId,
-                  type: item.type,
+              child: LockedFocusRow<AggregatedItem>(
+                key: _rowKeys[type],
+                items: items,
+                hubKey: 'favorites_row_${type.name}',
+                controller: scrollController,
+                height: rowHeight,
+                itemExtent: imageH * ar,
+                itemSpacing: 12 * desktopScale,
+                padding: EdgeInsets.fromLTRB(
+                  20 * desktopScale,
+                  5 * desktopScale,
+                  20 * desktopScale,
+                  5 * desktopScale,
                 ),
-              ),
-              itemBuilder: (context, item, index, isFocused) {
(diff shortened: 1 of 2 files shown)


Brady3035:
I also fixed the home rows to load items scrolled with trackpad or touch controls with that second commit

lib/ui/screens/browse/favorites_screen.dart
```diff
@@ -364,82 +364,94 @@ class _FavoritesScreenState extends State<FavoritesScreen> with GridFocusNodeMix
...
+                  final maxScroll = notification.metrics.maxScrollExtent;
+                  final currentScroll = notification.metrics.pixels;
+                  
+                  // 2. Trigger a load when the user scrolls within 500 pixels of the edge
+                  if (maxScroll - currentScroll < 500) {
+                    _vm.loadMoreForType(type);
+                  }
                 }
+                return false; 
               },
```
RadicalMuffinMan, line 380:
```suggestion
              onNotification: (notification) {
                // Page in more when a horizontal scroll gets within 600px of the end.
                if (notification.metrics.axis == Axis.horizontal &&
                    notification.metrics.maxScrollExtent -
                            notification.metrics.pixels <
                        600) {
                  _vm.loadMoreForType(type);
                }
                return false;
              },
```

Tidied up here. There were two trailing whitespace spots (the blank line in the middle and the space after return false), and the threshold is now 600 to match the home screen instead of 500, so both screens page at the same distance. loadMoreForType guards against duplicates internally, so firing on every scroll frame in the zone is fine.

lib/ui/screens/browse/favorites_screen.dart
```diff
@@ -364,82 +364,94 @@ class _FavoritesScreenState extends State<FavoritesScreen> with GridFocusNodeMix
...
-                return MediaCard(
-                  title: item.name,
-                  subtitle: _cardSubtitle(item),
-                  imageUrl: _imageUrl(
-                    item,
-                    maxWidth: width.toInt(),
-                    maxHeight: imageH.toInt(),
+                onIndexChanged: (index, item) {
+                  _onItemFocused(item);
+                },
```
RadicalMuffinMan, line 397:
```suggestion
                onIndexChanged: (index, item) {
                  _onItemFocused(item);
                  if (index >= items.length - 8) {
                    _vm.loadMoreForType(type);
                  }
                },
```

This restores the index based trigger you removed. The home screen keeps both its index trigger and the new scroll one, so this brings Favorites in line with that. The reason to keep it is that on TV the scroll trigger relies on the programmatic focus scroll actually landing within 600px of the end, whereas index >= length - 8 is an exact, layout-independent signal. `items` is in scope here (it's the same list passed to the row), and the view model dedups, so there is no downside to having both.

lib/ui/screens/home/home_screen.dart
```diff
@@ -1882,10 +1883,24 @@ class _ContentRowsState extends State<_ContentRows>
...
+
+  void _onRowScrolled(int rowIndex, ScrollController controller) {
+    if (!controller.hasClients) return;
+    const _loadMoreTriggerDistance = 600.0;
+    final remaining =
+        controller.position.maxScrollExtent - controller.offset;
+    if (remaining <= _loadMoreTriggerDistance) {
+      widget.viewModel.loadMoreForRow(rowIndex);
+    }
   }
```
RadicalMuffinMan, line 1904:
```suggestion
  void _onRowScrolled(int rowIndex, ScrollController controller) {
    if (!controller.hasClients) return;
    const loadMoreTriggerDistance = 600.0;
    final remaining =
        controller.position.maxScrollExtent - controller.offset;
    if (remaining <= loadMoreTriggerDistance) {
      widget.viewModel.loadMoreForRow(rowIndex);
    }
  }
```

Small naming fix. The underscore prefix is the convention for library private members, not local variables, so loadMoreTriggerDistance reads better here. The 600 already matches what Favorites now uses after the change above. This sits next to the existing index-based loadMoreForRow trigger and both deduplicate, so they coexist fine.

RadicalMuffinMan (changes requested):

Brady3035:
@RadicalMuffinMan why did this get closed did you merge the updated changes somewhere else?

RadicalMuffinMan:
@Brady3035 this was closed by accident when I deleted the patch branch, it wasnt intentional. can you resubmit please?

PR #636, opened 2026-06-24
Add setting to group library items into collections client-side

### LINKED ISSUE #622: [Feature] Group Items into Collections

I prefer for my movies and shows to be grouped into collections when browsing those respective libraries, but Moonfin does not currently allow that, even if the server-side settings are checked yes.

### Proposed solution

Would it be possible for the app(s) to respect the server-side setting and display movies or shows as grouped beneath their collections in those respective libraries?

### Platform

All / Not platform-specific

### Alternatives considered

_No response_

### Additional context

_No response_

mattsigal:
# Pull Request

## Summary
This PR adds a client-side library personalization setting: "Group Items into Collections". When enabled, it groups movies and shows that belong to collections under their respective boxsets (displayed alphabetically in the library view) and hides the individual movie or show items.

## Related Issues

- Closes #622 
- Fixes #
- Related to #

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Performance improvement
- [X] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

- **User Preference definition**: Added `groupItemsIntoCollections` preference key in `UserPreferences`.
- **Settings Screen Integration**:
  - Added a `SwitchPreferenceTile` in the Libraries section of `SettingsSidePanel`.
  - Added localization strings in `app_en.arb`.
  - Added a popup notice when toggling the setting ON if the server-side grouping options (`EnableGroupingMoviesIntoCollections` and `EnableGroupingShowsIntoCollections`) are not active on the server.
- **API and Endpoint adjustment**: Changed the `/Items` endpoint to `/Users/$userId/Items` in `JellyfinItemsApi` to correctly support user-scoped query filtering, specifically respecting the `CollapseBoxSetItems` parameter.
- **ViewModel adjustment**:
  - Listened to preference changes in `LibraryBrowseViewModel` to trigger page reload automatically.
  - Dynamically altered queried `includeTypes`, `excludeTypes`, and `collapseBoxSets` when requesting movies or shows if the preference is enabled.

## Platform
- [ ] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code

## Testing

### Test Steps
1. Navigate to Settings -> Personalization -> Libraries and toggle "Group Items into Collections" ON.
2. Verify the warning dialog displays indicating server settings requirement triggers if the server setting is OFF.
3. Turn the server settings ON and toggle "Group Items into Collections" ON.
4. Browse movies and TV shows libraries alphabetically to verify collections collapse and individual items are hidden.
5. Verify toggling setting OFF returns the display to individual items list.

* New Setting:
Shield_Screenshot_2026-06-24_12-18-51

* Ungrouped Library
Shield_Screenshot_2026-06-24_12-19-07

* Movie Collection Library
Shield_Screenshot_2026-06-24_12-18-27

* TV Show Collection Library
Shield_Screenshot_2026-06-24_12-34-30

* Warning Message about Server-side Toggle:
Shield_Screenshot_2026-06-24_12-19-35

## Checklist

### DIFF
--- a/lib/data/viewmodels/library_browse_view_model.dart
+++ b/lib/data/viewmodels/library_browse_view_model.dart
@@ -310,12 +310,10 @@ class LibraryBrowseViewModel extends ChangeNotifier {
     final groupCollections = _prefs.get(UserPreferences.groupItemsIntoCollections);
     if (includeItemTypes != null) {
       includeTypes = List<String>.from(includeItemTypes!);
-      if (groupCollections) {
-        if (includeTypes.contains('Movie')) {
-          collapseBoxSets = true;
-        } else if (includeTypes.contains('Series')) {
-          collapseBoxSets = true;
-        }
+      if (groupCollections &&
+          (includeTypes.contains('Movie') ||
+              includeTypes.contains('Series'))) {
+        collapseBoxSets = true;
       }
     } else {
       switch (_collectionType) {
--- a/lib/data/viewmodels/library_browse_view_model.dart
+++ b/lib/data/viewmodels/library_browse_view_model.dart
@@ -89,6 +89,8 @@ class LibraryBrowseViewModel extends ChangeNotifier {
   String? _errorMessage;
   String? get errorMessage => _errorMessage;
 
+  bool? _lastGroupCollectionsValue;
+
   AggregatedItem? _focusedItem;
   AggregatedItem? get focusedItem => _focusedItem;
 
@@ -170,6 +172,16 @@ class LibraryBrowseViewModel extends ChangeNotifier {
     _letterFilter = '';
     _imageType = _prefs.get(UserPreferences.libraryImageType(_imagePrefKey));
     _posterSize = _readScopedPosterSize();
+    _lastGroupCollectionsValue = _prefs.get(UserPreferences.groupItemsIntoCollections);
+    _prefs.addListener(_onPrefsChanged);
+  }
+
+  void _onPrefsChanged() {
+    final newValue = _prefs.get(UserPreferences.groupItemsIntoCollections);
+    if (_lastGroupCollectionsValue != newValue) {
+      _lastGroupCollectionsValue = newValue;
+      load();
+    }
   }
 
   String get _prefKey => genreId ?? libraryId;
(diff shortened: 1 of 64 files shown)


lib/data/viewmodels/library_browse_view_model.dart
```diff
@@ -295,18 +307,37 @@ class LibraryBrowseViewModel extends ChangeNotifier {
...
     if (includeItemTypes != null) {
-      includeTypes = includeItemTypes;
+      includeTypes = List<String>.from(includeItemTypes!);
+      if (groupCollections) {
+        if (includeTypes.contains('Movie')) {
+          collapseBoxSets = true;
+        } else if (includeTypes.contains('Series')) {
+          collapseBoxSets = true;
+        }
+      }
```
RadicalMuffinMan:
```suggestion
      if (groupCollections &&
          (includeTypes.contains('Movie') ||
              includeTypes.contains('Series'))) {
        collapseBoxSets = true;
      }
```

The if and else if both set collapseBoxSets = true, so they collapse into one condition. Same behavior, less nesting.

lib/ui/screens/settings/panel/libraries_category_screen.dart
```diff
@@ -33,6 +33,46 @@ class _LibrariesCategoryScreen extends StatelessWidget {
...
+                    if (client.serverType == ServerType.jellyfin) {
+                      try {
+                        final config = await client.adminSystemApi.getServerConfiguration();
+                        final groupMovies = config['EnableGroupingMoviesIntoCollections'] as bool? ?? false;
+                        final groupShows = config['EnableGroupingShowsIntoCollections'] as bool? ?? false;
+                        if (groupMovies && groupShows) {
+                          return;
+                        }
+                      } catch (_) {}
+                    }
```
RadicalMuffinMan:
```suggestion
                    final client = GetIt.instance<MediaServerClient>();
                    try {
                      final config =
                          await client.adminSystemApi.getServerConfiguration();
                      final groupingEnabled =
                          config['EnableGroupingIntoCollections'] as bool? ??
                              false;
                      if (groupingEnabled) {
                        return;
                      }
                    } catch (_) {}
```

Two fixes in one. First, dropping the serverType == jellyfin guard so the check runs on Emby too. Second, Jellyfin's ServerConfiguration has no EnableGroupingMoviesIntoCollections/...Shows, the real setting is a single EnableGroupingIntoCollections. The old code read two non-existent keys, so they were always null and the warning fired even when the server was set up correctly. This reads the real key. If an Emby server happens to use a different key, the ?? false just falls back to showing the warning, which is harmless.

RadicalMuffinMan (changes requested):

lib/ui/screens/settings/panel/libraries_category_screen.dart
```diff
@@ -33,6 +33,46 @@ class _LibrariesCategoryScreen extends StatelessWidget {
...
+                    final client = GetIt.instance<MediaServerClient>();
+                    try {
+                      final config =
+                          await client.adminSystemApi.getServerConfiguration();
+                      final groupingEnabled =
+                          config['EnableGroupingIntoCollections'] as bool? ??
+                              false;
+                      if (groupingEnabled) {
+                        return;
+                      }
```
RadicalMuffinMan:
```suggestion
                      final groupMovies =
                          (config['EnableGroupingMoviesIntoCollections'] as bool?) ??
                          (config['EnableGroupingIntoCollections'] as bool?) ??
                          false;
                      final groupShows =
                          (config['EnableGroupingShowsIntoCollections'] as bool?) ??
                          (config['EnableGroupingIntoCollections'] as bool?) ??
                          false;
                      if (groupMovies && groupShows) {
                        return;
                      }
```

The single EnableGroupingIntoCollections key only exists on Jellyfin 10.10. On 10.11 and 12.0 it was split into EnableGroupingMoviesIntoCollections and EnableGroupingShowsIntoCollections, so on those servers the single key reads null, falls to ?? false, and the warning fires even when grouping is correctly enabled. This reads the two split keys first and falls back to the old single key, so it covers 10.10.7 through 12.0. Suppressing only when both movie and show grouping are on keeps it conservative, since a library could be either type. You may want to run dart format on this, the fallback lines run a little long.

RadicalMuffinMan (changes requested):

PR #662, opened 2026-06-28
Feature/fix - tweak preference initilization

broken-droid:
## Summary
Rework initialization of language related settings from server.  Fixed some preferences being global when they should have been user scoped, which caused problems if there were multiple users.  Initialized the new subtitlemode setting to be mapped from the users jellyfin subtitle mode setting.

## Related Issues

- Closes #
- Fixes #
- Related to #

## Type of Change
- [X] Bug fix
- [X] New feature
- [X] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

- added subtitleMode handling to system_models.dart, and map server's json enum to a string if it uses ints
- add subtitleMode handling to initLanguagePrefs, and map jellyfin style enum strings to moonfin's
- add missing prefs to the _scopedPreferenceKeys:
pref_subtitle_mode
pref_fallback_audio_language
pref_prefer_default_audio_track
pref_prefer_audio_description
pref_fallback_subtitle_language
- add helper _setIfMissing - saves a scoped pref if it doesn't exist, used this in initLanguagePrefs so it correctly saves user scoped prefs now
- updated containsPreference to add a 'scopedOnly' flag
- updated initLanguagePrefs, resolveLibraryPosterSize, and resolvePlaylistPosterSize to use containsPreference with scopedOnly=true
- remove _migrateSubtitleModePreference - handled by new init logic
- remove _initializeFallbackSubtitleLanguagePreferences - not needed since accessor has a default value
- updated the set method's subtitleMode logic to update scoped subtitle language preferences instead of global

## Platform
- [ ] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code

## Testing

### Test Steps
1. create a user with foreign audio and subtitle languages, and a subtitle mode
2. log in with that user
3. create a different user with different audio/subtitle/mode settings than the previous
4. log in and verify they are correct

## Checklist


RadicalMuffinMan:
good catch thank you

PR #675, opened 2026-06-29
Refresh home after Seerr bootstrap

matschi95:
# Pull Request

## Summary

Fixes a startup timing issue where Seerr home rows configured through the plugin-backed Home Sections layout only appeared after opening the Home Row settings screen.

Home now reacts to Seerr availability changes directly, and Seerr Home Sections are gated by `PluginSyncService.seerrAvailable` instead of the legacy `SeerrPreferences.rowsConfig` row configuration.

## Related Issues

None

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

- Added a small `HomeRefreshBus` abstraction for immediate and pending Home refresh requests.
- Refresh Home after Seerr bootstrap when the Seerr repository becomes available.
- Let `HomeShell` listen for `PluginSyncService.seerrAvailable` changes and refresh Home when that value changes.
- Use `PluginSyncService.seerrAvailable` as the global Seerr visibility gate for Home Sections.
- Removed legacy `SeerrPreferences.rowsConfig` checks from Home and Home Row settings visibility decisions.
- Kept Home Section activation and ordering owned by `HomeSectionConfig`, including plugin-provided defaults.

## Platform
- [x] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [x] Linux
- [x] All / Shared code

## Testing

### Test Steps

1. Configure Seerr Home Sections through the server plugin defaults.
2. Start the app fresh and open the Home screen without opening settings first.
3. Verify configured Seerr rows appear after Seerr availability is established.
4. Open Home Row settings and verify plugin-provided Collections are not duplicated.
5. Run targeted analyzer checks for the changed files.

None

## Checklist

### DIFF
--- a/lib/data/services/plugin_sync_service.dart
+++ b/lib/data/services/plugin_sync_service.dart
@@ -510,7 +510,7 @@ class PluginSyncService extends ChangeNotifier {
         password: password,
       );
       await _refreshAvailabilityStatus(client);
-      return seerrRepo.isAvailable;
+      return seerrAvailable;
     } catch (_) {
       return false;
     }
--- a/lib/data/services/plugin_sync_service.dart
+++ b/lib/data/services/plugin_sync_service.dart
@@ -491,15 +491,15 @@ class PluginSyncService extends ChangeNotifier {
     }
   }
 
-  Future<void> configureSeerr(
+  Future<bool> configureSeerr(
     MediaServerClient client, {
     String? username,
     String? password,
   }) async {
-    if (!_pluginAvailable) return;
+    if (!_pluginAvailable) return false;
 
     final token = client.accessToken;
-    if (token == null || token.isEmpty) return;
+    if (token == null || token.isEmpty) return false;
 
     try {
       final seerrRepo = await GetIt.instance.getAsync<SeerrRepository>();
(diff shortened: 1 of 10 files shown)


lib/data/services/plugin_sync_service.dart
```diff
@@ -509,7 +509,11 @@ class PluginSyncService extends ChangeNotifier {
         username: username,
         password: password,
       );
-    } catch (_) {}
+      await _refreshAvailabilityStatus(client);
+      return seerrRepo.isAvailable;
```
RadicalMuffinMan:
```suggestion
      await _refreshAvailabilityStatus(client);
      return seerrAvailable;
```
small thing on the return here. this returns seerrRepo.isAvailable, which is status.authenticated, so it only goes true once the Seerr SSO actually logs in. but Home decides whether to show Seerr rows based on PluginSyncService.seerrAvailable, which is just pluginAvailable and seerrEnabled and the user pref, it doesn't care about auth.

session_repository uses this return value to fire the home refresh on login. so if Seerr is enabled but the SSO handshake hasn't finished or fails, the gate says show the rows but the trigger never fires, and you end up back at the same missing-rows-on-startup bug this PR is fixing, just for that case.

simplest fix is to return seerrAvailable instead so the refresh trigger keys off the exact same signal Home gates on. the getter is on this class and refreshAvailabilityStatus right above already updated the state it reads, so it reflects the right value here.

RadicalMuffinMan (changes requested):

lib/data/services/plugin_sync_service.dart
matschi95: done :)

PR #694, opened 2026-07-01
New Sort Option: Original

matschi95:
# Pull Request

## Summary
Adds an `Original` sort option for configurable Home rows. When selected, Moonfin preserves the server-provided/default item order by omitting the `sortBy` query parameter instead of applying a Jellyfin sort field.

Home row sort preferences now support `Original` for Favorites, Collections, Genres, Playlists, and Audio playlists. These Home rows also default to `Original`, so their natural order is preserved unless the user explicitly chooses another sort mode.

Normal browse/library sorting keeps the existing behavior and still defaults to sorting by name.

## Related Issues
None.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Performance improvement
- [x] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
- Added `LibrarySortBy.original`.
- Added nullable sort parameter mapping so `Original` resolves to no `sortBy` argument.
- Exposed `Original` for Home row sort preferences while keeping regular browse/library sort options unchanged.
- Changed configurable Home row sort defaults to `Original`.
- Updated Home row loading paths so `sortBy` is only passed when the selected option is not `Original`.
- Applied the same behavior across single-server and multi-server row loading.

## Platform
- [x] Android
- [x] iOS
- [x] tvOS
- [x] Web
- [x] macOS
- [x] Windows
- [x] Linux
- [x] All / Shared code

## Testing
- [ ] Not tested (explain why)

### Test Steps
changed Home-Collection-Ordering from `Name` to `Original` and back to `Name`

## Checklist

### DIFF
--- a/lib/data/repositories/multi_server_repository.dart
+++ b/lib/data/repositories/multi_server_repository.dart
@@ -59,6 +59,9 @@ class MultiServerRepository {
   static const _defaultSortOrder = 'Ascending';
   static const _genreArtworkConcurrency = 6;
 
+  static String? _sortByParameter(LibrarySortBy? sortBy) =>
+      sortBy == null ? _defaultSortBy : sortBy.sortByParameter;
+
   List<ServerUserSession>? _cachedSessions;
   DateTime _cacheExpiry = DateTime(0);
   final Map<String, int> _rowOffsets = {};
@@ -277,8 +280,8 @@ class MultiServerRepository {
 
   Future<HomeRow> getAggregatedPlaylists({
     int limit = _defaultLimit,
-    String sortBy = _defaultSortBy,
-    String sortOrder = _defaultSortOrder,
+    String? sortBy = _defaultSortBy,
+    String? sortOrder = _defaultSortOrder,
     String? mediaType,
   }) async {
     final sessions = await getLoggedInServers();
@@ -314,14 +317,20 @@ class MultiServerRepository {
     );
 
     final all = results.expand((e) => e).toList();
-    if (sortBy == 'SortName') {
+    if (sortBy == null) {
+      // Keep the server-provided order.
+    } else if (sortBy == 'SortName') {
       if (sortOrder == 'Ascending') {
         all.sort((a, b) => a.name.compareTo(b.name));
       } else {
         all.sort((a, b) => b.name.compareTo(a.name));
       }
     } else {
-      _sortAggregatedItems(all, sortBy: sortBy, sortOrder: sortOrder);
+      _sortAggregatedItems(
+        all,
+        sortBy: sortBy,
+        sortOrder: sortOrder ?? _defaultSortOrder,
+      );
     }
 
     final takenItems = all.take(limit).toList();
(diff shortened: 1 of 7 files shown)


RadicalMuffinMan:
Please use the provided PR template

matschi95:
I have changed to the template (sorry for that).

This PR is essential for the implementation of automatic Trending Collections (via the CollectionImportPlugin), as otherwise the ranking gets ruined by the sorting.

RadicalMuffinMan:
CollectionImportPlugin?

matschi95:
This plugin (https://github.com/lostb1t/jellyfin-plugin-collection-import) allows you to dynamically generate collections based on mdblist (and have them updated daily). For example, a ‘Trending Movies’ collection. You can then pin this collection to your home screen to display a row showing the currently trending films. The order within the collection should not be altered, as it reflects the ranking (therefore `original` sort order).

mattsigal:
That's not needed with the upcoming Home Rows feature, where we support TMDb lists on the home screen natively:

2026-07-01_11-25-56_mspaint

matschi95:
Wow, that looks great. But unfortunately, you’re limited to just a few lists there. I like to have lots of custom lists on my home screen, such as ones from mlotek (https://mdblist.com/lists/_mlotek/). Or if you want to create a Marvel list in chronological order. There’s definitely a point to having an original order.

mattsigal:
> Wow, that looks great. But unfortunately, you’re limited to just a few lists there. I like to have lots of custom lists on my home screen, such as ones from mlotek (https://mdblist.com/lists/_mlotek/). Or if you want to create a Marvel list in chronological order. There’s definitely a point to having an original order.

Actually, we have full custom list support coming in the same update. The ordering is maybe an issue (haven't really looked at that particular nuance but we have options for it): 

2026-07-01_11-42-43_moonfin

2026-07-01_11-42-50_moonfin

I would suggest closing this PR until after 2.3 drops and then revisiting it if needed.

matschi95:
That feature looks really cool. Are these custom rows something the plugin can create for all clients as well?

Also, this feature doesn’t fully solve the issue. It seems like a SortBy setting is still required, and that disrupts the ordering of a ranked collection completely.

mattsigal:
Yeah, that why I said earlier that the sorting option _might_ still be needed although I'm almost positive that the default sort option does what you want (eg. it just puts them in the order received from TMDb, so the first item on their list is the leftmost on the Home Row). Maybe it should say Default (list order) instead of (when added), as that really is what it is doing.

This is a brand new feature, so for the moment the rows are being defined within the client. Creating them within Moonbase sounds like an interesting idea and might be something worth making a feature request for once the feature is live and people get a chance to use it.

RadicalMuffinMan:
Thank you for putting this together, and sorry it sat open so long.

Good news is this is already solved in 2.4.0. We added a "Playlist Order" sort option that skips the server sort entirely, and Collections and Playlists home rows now default to it. So a pinned collection from the CollectionImportPlugin should keep its MDBList ranking out of the box, without changing any setting.

PR #772, opened 2026-07-11
Playback Tweaks to address connection abort errors on pause/resume and theme music overlap

mattsigal:
# Pull Request

## Summary
This pull request fixes two playback-related issues on the desktop clients:
1. TCP connection aborts (`WSAECONNABORTED`/`10053` and `WSAECONNRESET`/`10054`) causing playback stalls or spurious error banners after leaving video paused for 5-10 minutes.
2. Background show theme music playing when active video is paused inside the player.

## Related Issues

Discord feedback and in-player experience.

- Closes #
- Fixes #
- Related to #

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

- Added low-level TCP/HTTP auto-reconnection properties (`reconnect`, `reconnect_on_network_error`, `reconnect_on_http_error`) to libmpv's `NativePlayer` config in `MediaKitPlayerBackend`.
- Filtered transient protocol-level warnings (prefixed with `tcp:`, `http:`, `https:`, `tls:`, `crypto:`) in `MediaKitPlayerBackend.errorStream` to prevent them from bubbling up to `PlaybackManager` as fatal player errors on recovery when they are transient issues that are resolved by the time they fire.
- Updated `ThemeMusicService.playForItem` to return early if any media item (audio or video) is active in `PlaybackManager.queueService.currentItem`. This ensures details-page theme music remains quiet while the player is open/active.

## Platform
- [ ] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code

## Testing

- [X] Manual testing completed - testing on Windows desktop. Flutter run testing was completely clean.

### Test Steps
1. Play a movie/show episode, pause it, and wait 10 minutes for the TCP socket to expire.
2. Resume playback and verify that it resumes cleanly without any error banners.
3. Pause and play the media multiple times inside the player and verify that background theme music does not trigger.

## Checklist

### DIFF
--- a/lib/data/services/theme_music_service.dart
+++ b/lib/data/services/theme_music_service.dart
@@ -73,18 +73,7 @@ class ThemeMusicService {
   Future<void> playForItem(AggregatedItem item) async {
     if (!_prefs.get(UserPreferences.themeMusicEnabled)) return;
     final activeItem = _playbackManager.queueService.currentItem;
-    bool isAudio = false;
-    if (activeItem is AggregatedItem) {
-      isAudio = activeItem.isAudioLike;
-    } else if (activeItem is String) {
-      final meta = _playbackManager.currentOfflineMetadata;
-      if (meta != null) {
-        final type = meta['Type']?.toString();
-        final mediaType = meta['MediaType']?.toString();
-        isAudio = type == 'Audio' || type == 'AudioBook' || mediaType == 'Audio';
-      }
-    }
-    if (isAudio) return;
+    if (activeItem != null) return;
     if (_playbackManager.state.isPlaying) return;
     if (_externalAudioActive) return;
     if (_playSuppressed) return;
--- a/lib/playback/media_kit_player_backend.dart
+++ b/lib/playback/media_kit_player_backend.dart
@@ -378,6 +378,11 @@ class MediaKitPlayerBackend extends PlayerBackend {
     final platform = player.platform;
     if (platform is NativePlayer) {
       _nativeSetProperty(platform, 'network-timeout', '120');
+      _nativeSetProperty(
+        platform,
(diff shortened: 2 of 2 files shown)


lib/data/services/theme_music_service.dart
mattsigal, line 1: Note that the isAudio change is because we want any media (audio OR video) that is active in the player's queue to block the theme music. Since the active item stays in the queue even when a movie is paused, the new check blocks the theme music for both paused videos and background music. This made the previous type-checking/casing lines for isAudio redundant, so I believe can be safely removed.

RadicalMuffinMan:
The reconnect stuff is great for the pause/resume drops, but I'd yank two of those options.

reconnect_at_eof=1 makes ffmpeg try to reconnect every time it hits the end of a stream. Good for live TV but on a normal movie the end of the file is just the end, so it'll sit there reconnecting at the finish of every video and can mess with autoplay-next or hang on the last second.

reconnect_on_http_error=4xx,5xx retries on stuff like 401/403/404 too, which is pointless. If the token's expired or the stream's gone, retrying just spams the server. The network-error reconnect already covers the real case.

So just trim it to:
```
reconnect=1,reconnect_on_network_error=1,reconnect_streamed=1,reconnect_delay_max=5
```
Still fixes the paused-socket-reset thing, still reconnects for transcodes, caps the backoff at 5s, and leaves normal endings and real auth fails alone.

mattsigal:
Good points, I'll update soon.

PR #872, opened 2026-07-20
feat(games): add game system/games "library" navigation heirarchy

WizardOfYendor1:
# Pull Request

## Summary
Provide a brief description of what this PR changes and why.

## Type of Change
- [ ] Bug fix
- [X ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ X] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

For the recent retro game feature added (thanks, really cool!), I've tweaked the retro ROM library page and navigation. 
With multiple systems each containing hundreds of ROMs, it's bit of a pain to navigate. I added a root screen, which is a game system level view, that you can then drill down into to search and vertically scroll through the game list and search (using Moonfin's look/feel).  It searches on  word prefixes so if for example you search for "River", it returns "River Raid" and not "Night Driver" games. On my disk/library I have folders that are organized by system then the ROMs are located inside those folders. 

I mostly replicated the functionality of the Moonfin library navigation...which means cut/pasting/duplication of code.  I also "respected" Moonfin's poster-size and scale user preferences. 

I didn't try to extract widgets like the alphabet sorter out into a shared widget because I thought it too ambious/dangerous to be refactoring all of that unrelated library code. I did however make a shared widget for this games implementation which could be used as a template for a future refactoring effort. 

I should also add, I didn't do any i18n with my change(s). There's existing need from the base implementation that - so perhaps this all should be one PR to deal with it? 

Screenshots below. I'll not be offended if this gets totally refactored/gutted as I am still getting familiar with the code base. 

Please give extra scrutiny as this is my 1st PR into the Moonfin world and probably is a little more ambitious than I should've taken on for my 1st one :-). 

## Platform
- [ ] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ X] All / Shared code

## Testing
I've tested the web client, android and android TV.  I don't have any Apple "ecosystem" devices (or emulators setup) so it's untested there and other devices like Roku et al

- [X] Not tested (explain why): Some devices like Apple or Roku I don't have physically or emulators 

### Test Steps
1. Setup the retro game llibrary
2. Open the retro library via Moonfin
3. Drill into each system (which is the parent folder on disk)

image

image

image

## Checklist
- [X ] Code builds successfully
- [ X] Code follows project style and conventions
- [X ] No unnecessary commented-out code
- [X ] No new warnings introduced

### DIFF
--- a/lib/data/viewmodels/game_system_browse_view_model.dart
+++ b/lib/data/viewmodels/game_system_browse_view_model.dart
@@ -1,0 +1,275 @@
+import 'dart:async';
+
+import 'package:flutter/foundation.dart';
+import 'package:server_core/server_core.dart';
+
+import '../../util/game_browse_filter.dart';
+import '../../util/game_library.dart';
+
+enum GameSystemBrowseError { unsupported, loadFailed }
+
+/// Games-scoped state for browsing one retro-game system.
+///
+/// Focus nodes and responsive layout remain in the screen; API work, prepared
+/// search data, debounce timers, selection, and detail caching live here so the
+/// widget does not mix rendering with data orchestration.
+class GameSystemBrowseViewModel extends ChangeNotifier {
+  GameSystemBrowseViewModel({
+    required GamesApi? gamesApi,
+    required this.libraryId,
+    required this.systemId,
+    String? systemName,
+  }) : _gamesApi = gamesApi,
+       _providedSystemName = systemName?.trim();
+
+  static const searchDebounceDuration = Duration(milliseconds: 150);
+  // Descriptions should feel immediate while still ignoring a pointer merely
+  // crossing a card. Full-screen backdrop work can wait until navigation has
+  // clearly settled so normal browsing does not churn blurred backgrounds.
+  static const detailDebounceDuration = Duration(milliseconds: 75);
+  static const backdropDebounceDuration = Duration(milliseconds: 650);
+
+  final GamesApi? _gamesApi;
+  final String libraryId;
+  final String systemId;
+  final String? _providedSystemName;
+
+  bool loading = true;
+  GameSystemBrowseError? error;
+  String selectedLetter = '';
+  String appliedSearchQuery = '';
+  String? resolvedSystemName;
+  List<GameSummary> games = const [];
+  List<GameSummary> visibleGames = const [];
+  Map<String, String> displayTitles = const {};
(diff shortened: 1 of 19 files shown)


RadicalMuffinMan:
Hi! Currently games do show up in regular search results as well. The current games library layout was more of a v1 effort, I did have plans to restyle it akin to the search results with tabs but I really like your take on how its styled after the genres screen. I'd say take out the in-library search and instead show the game description at the top like other libraries do once the item is selected / hovered / focused.

WizardOfYendor1:
> Hi! Currently games do show up in regular search results as well. The current games library layout was more of a v1 effort, I did have plans to restyle it akin to the search results with tabs but I really like your take on how its styled after the genres screen. I'd say take out the in-library search and instead show the game description at the top like other libraries do once the item is selected / hovered / focused.

I kinda went off the deep end on this and turned out to be more of an adventure than I bargained for :-). On the plus side, I now have a greater understanding how the library display works :-)

After making the UI changes you suggested, sans the search bar removal, the artwork was loading a bit sluggish on androidtv. So I started down an ill-fated effort to create a dedicated game art/meta cache for multi-thousand ROM scenarios (most common). Long story short, this is where I landed. This will give priority loading of artwork of the focused row and load 5 rows above/below (in ordinal order 1, -1 etc) in the background. So if your scrolling through your 1000 game NES system, it's a bit snappier IMHO. I left the search bar because I think it's useful or perhaps I didn't understand what you meant. 

Ideally, the artwork delivered via the plugin should provide a thumbnail of a "requested" size (i.e a "maxWidth"). Probably easier said than done as Jellyfin handles this for media content it "knows" about (perhaps this can be leveraged for the ROM art?).  Adding game description to the lightweight "HUD" call would also help with snappiness. i.e. save the "big call" when you actually click into the game.

I also have concerns with multi-thousand ROM libraries and their artwork caching . The persistent cache can be dominated by retro rom artwork causing eviction of movie artwork (etc). If I've read this right, it's capped at 600 items and a few hundred MB. I have like 5000 ROMs total, and that's "paired down" :-). I think have this many+ is common. So perusing your ROM library could blow out your movie/series posters cache (IF I understand this right)


Testing on other devices like Apple stuff is REALLY necessary - esp the navigation.

PR #895, opened 2026-07-21
refactor(server): share Jellyfin/Emby API implementations via ServerDialect

MathiasDevelopes:
The Jellyfin and Emby API implementations were ~90% copy-pasted (~1,900 duplicated lines across 24 files). Both servers speak the same API with small, enumerable differences.

Behavior unifications (previously drifted copies):
- 5-min caches for user views and display preferences now apply to Jellyfin too (Emby already had them).
- Ancestors and special-features lookups fail gracefully on both.
- Jellyfin image/playback URLs resolve baseUrl lazily, fixing stale URLs after a server URL change.
- Image query params are URL-encoded on both servers.
- Emby now logs network requests/responses/errors like Jellyfin already did (shared attachServerInterceptors).

Tests: twin playback/live_tv test pairs merged into server_core, plus a new dialect test pinning per-server paths and capability gates.

# Pull Request

## Summary
This PR deduplicates Jellyfin/Emby API logic, by consolidating common logic and storing the differences in a `ServerDialect` class.

## AI Assistance
Refactor plan and implementation were generated by Kimi K3 (Moonshot AI).
I spot-checked key files rather than reviewing the full diff line-by-line,
and tested the build against a Jellyfin and Emby server (see Test Steps).

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
- Add ServerDialect capturing auth scheme, user-scoped vs user-less endpoint paths, capability flags (QuickConnect, lyrics, media segments, remote subtitle search), api_key image auth and the per-server display-preferences client name.
- Move the 12 duplicated API implementations into server_core as dialect-parameterized Server*Api classes.
- Extract shared interceptor wiring (auth header, redirects, network logging) into attachServerInterceptors.
- Both media server clients are now thin wiring shells; package names and client classes unchanged, so app code is untouched.

## Platform
- [ ] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [x] All / Shared code

## Testing

### Test Steps
Tested Linux build against Jellyfin and Emby server, verified that login, library browsing, playback, favorites, and resume working.

## Checklist


RadicalMuffinMan:
Thanks for taking this on and sorry it sat open so long.

I'm going to close it. The duplication you spotted is real, but the premise that the two servers differ in small enumerable ways hasn't held up. Since July, main has landed five commits that exist only because Emby behaves differently: hidden library excludes (#1022), chapter markers read two ways and intro binding (#1007), a field name Emby rejects (#1094), and remote subtitle search (#1162). All of them landed in the two files this PR removes.

Some of those differences aren't flag-shaped either. getMediaSegments on Emby is now a translation layer that reads chapter markers and infers a missing intro end from the following chapter. There's no Jellyfin equivalent
to share it with.

That drift has already caught the dialect out. ServerDialect.emby declares supportsRemoteSubtitleSearch: false, but Emby has had it since #1162, so merging this would quietly switch a shipped feature back off with nothing failing to compile. supportsMediaSegments has the same problem.

The other thing giving me pause is scale. This removes 1864 lines across the whole server API layer, it's conflicting against a month of new work in those same files, and any mistake lands on both servers at once. It also folds in
real behaviour changes, caching, error handling, lazy baseUrl resolution and URL encoding, which deserve their own review rather than riding inside a refactor this size.

None of that is wasted effort though. Two pieces here would be very welcome as small standalone PRs:

- attachServerInterceptors, plus the Emby request logging that comes with it
- sharing the APIs that really are near-identical, one at a time, each with its own tests. system_api, live_tv_api, playback_api, instant_mix_api and image_api are within a few lines of each other on both sides

That's most of the win with a diff we can actually review. Thanks again for the writeup, and for being upfront about how it was put together.

PR #989, opened 2026-07-30
Fix elevated HDR black levels on tvOS native playback

harveyplath:
## Summary

- derive the sample format description from the decoded `CVPixelBuffer`
- stop reusing the compressed `hvc1`/`dvh1` input description for decoded `x420` frames
- preserve video-range and HDR colour attachments for `AVSampleBufferDisplayLayer`

## Reproduction

1. Direct Play a limited-range BT.2020/PQ HEVC Main10 black-level pattern on Apple TV.
2. Compare Moonfin with Swiftfin on the same display.
3. Moonfin renders nominal black grey after decoded frames begin; Swiftfin renders pitch black.

The source measures nominal 10-bit limited-range black (Y=64). Moonfin initially shows a pitch-black display-layer background, which becomes grey when decoded frames start, isolating the presentation path.

## Fix

`CMSampleBufferCreateReadyWithImageBuffer` now receives a `CMVideoFormatDescription` created from the decoded image buffer. The compressed description remains available for display criteria and source dimensions only.

## Validation

- `git diff --check` passes.
- Runtime verification requires a signed tvOS build and the controlled Y=64 HDR pattern.

### DIFF
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -114,6 +114,17 @@ android {
             manifestPlaceholders["appName"] = "$baseAppName Beta"
             manifestPlaceholders["enableImpeller"] = "false"
         }
+        // Deliberately a separate installable TV package for local builds. It
+        // must never replace the signed upstream Android TV application.
+        create("androidTvHarvey") {
+            dimension = "device"
+            applicationId = "$baseAppId.harvey"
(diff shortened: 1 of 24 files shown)


harveyplath:
Closing because the affected native tvOS playback implementation exists only on the personalized branch and is not present on current upstream main.

PR #998, opened 2026-08-01
fix(network): use a browser-compatible Moonfin user agent

gavinzhangth:
## Summary

- set a browser-compatible, Moonfin-identifying user agent for native server requests
- avoid 403 responses from reverse proxies and WAFs that reject Dart's default user agent
- add a local HTTP server regression test

## Context

Some otherwise reachable Jellyfin and Emby endpoints reject `Dart/<version> (dart:io)`. Moonfin then exhausts its connection probes before reporting that it cannot connect. Configuring the shared native `HttpClient` fixes discovery and subsequent API traffic while keeping Moonfin explicitly identifiable in server logs.

## Testing

- `dart test` (5 tests passed)
- `dart analyze` (no issues found)


jmawet:
Please use the PR template

PR #1071, opened 2026-08-07
(feat) Arcade/Retro game playback with native libretro cores, android -> native controller bridge and server-side artwork refactor (ET AL)

WizardOfYendor1:
# Pull Request

# Summary
Add arcade/native core enhancements to retro gaming.  Also fixes problems detected during development and performance enhancements. 
_Sorry again.... trying to be brief but this longer than my other sister PR :-).  Also sorry for the mega PR,  I couldn't split this in a logical way but tried with the commits :-)._ 

#  Changes
This started innocently a few weeks ago. I'd just finished the nav screen and wondered if I could get arcade ROMs playing on a native core. Coded up support, hit problems, fixed them, rinse-repeat. It's in reasonable shape now, so I'm contributing it back, grouped into as few logical commits as I could manage. There's alot that I uncovered along the way and the description would be a tome if I went through every detail. 

My guiding principles were: don't regress anything, watch performance/memory/disk so less powerful streaming boxes still behave/work, and above all keep this isolated — nothing here should break Moonfin's day job of watching/storing media.

## Arcade: 
FBNeo turned out to be the most capable arcade core I found, and as a bonus, handles plenty of non-"arcade" systems too. The server hashes known ROMs so it can recommend a native core — which sounds way simpler than it was — and hides cores that don't apply. That shows on the game screen, and you can click through to EmulatorJS if a game refuses to run natively. Your choice is saved and persisted per user. This is rather challenging given all the romsets out there. MAME is there of course, running inside emulatorJS (you can also force/choose FBNeo - it rarely works but...)

Some cores have no software renderer (N64, for instance), so those games won't run at all. Hardware rendering is its own project — not touching it here. :-)

Also I added some "safety" here to limit/prevent malicious ROMs impact. With the JNI/C layer - that got my radar focused on this.

## Input:
That all led into gamepad work; latency, mapping, and getting native cores to behave not just for arcade but for things like the 2600 et al. Ended up building a bridge (as such) between the two worlds. Lots and lots of coding, testing, and dusting off C/C++ skills. Plus a pile of small fixes testing shook out — e.g. confirmation gates before exiting a game, save/load improvements, etc. I had plenty of issues with button lag/latency with my gamepad (using a Logitech wireless rumblepad). This required the old trusty bitwise operations. While the "speed" improvement I was after was neglible, it did pack Gamepad state into one bitmask and hands to the core in a single atomic write. "Speed" is kinda one but more importantly a "diagonal move" can't split across two frames. That ALSO led me to run into a real race in the input latch — a read-then-exchange could write back a stale bit value and report a button released mid-press. I'm still not entirely satisfied here but not sure how solvable this problem in with this code (i.e. limitations of the device/android bridge)

I also attempted to handle multiple controllers but this is largely untested/vetted right now. 

## Native libretro C:
"While I was in there", I took a deep dive (with AI help) through the libretro JNI/C layer for memory safety and malicious-ROM hardening, plus found and fixed a video quality bug. I extended the native test suite to cover the changes and ran it clean under ASAN, UBSAN and TSAN (memory/thread/unexpected behavior safety). Might be worth wiring those into CI going forward — they're not currently. I also had to tweak CMakeLists locally to get tests to run under windows WSL (MS VC). I didn't commit that "dev convenience" change here.  
I should also note, great minds think alike. When I merged up from upstream, I found that you implemented some of the things I had done.... thus merge conflicts I had to resolve.  I merged changes together the way I thought made sense given your intent.

## Artwork: 
With 5,000+ ROM libraries, artwork speed and storage became a real problem. So the plugin now pulls art from libretro and shrinks it server-side. The client doesn't block waiting for the small thumbnail — it'll take the full-size image if that's all that exists, cached for only 2 hours, and picks up the thumbnail as soon as the server publishes it (then THAT is cached a year, immutable). Reconciliation runs at startup and again whenever a library scan completes. I ripped out the whole scrolling viewport idea. It added needless complexity for little gain. With the smaller thumbnails, it loads pretty fast, esp once the cache fills up. 
I spent a ridiculous and shameful amount of time on this, and may have over-engineered it, but it works well. :-)

## Misc:
I was tempted to turn on impeller for performance... but just too many reported issues and the impact could go beyond this. Left alone.

I went through many refactors: build it, didn't like something, hit a wall, rip it out, try again. Expect some sub-optimal code that needs refactoring. Roughly 200 unit tests came out of it, mostly because that was the fastest way to exercise and debug the difficult parts — and as a bonus they catch regressions from my constant refactoring.

## LLM usage:
I want to describe as best I can, because it's a mix and not my usual MO — this was experimental and I wasn't sure it would work at all, so I leaned harder on AI more than normal.

Early on it was for research. I wrote most of the foundational code myself, using the LLM for testing and feedback. At certain point of playing whack-a-mole, I decided that building a "retro test harness" was a faster and better idea. Agents then ran closed-loop against the web client, Android TV emulators, and real androidtv hardware (via ADB) — pressing buttons, reviewing screens/functionality, reading logcat, iterating that in real time - all at my direction. Some parts I was very hands-on; others I scoped to an agent and reviewed. The unit tests were almost entirely agent-written against implementations and use cases I specified — so kinda TDD-ish. 

##
Two weeks, more hours than I want to count, and it "works on my machine" — meaning Windows and AndroidTV, with my ROM set. iOS and other ROM sets genuinely need testing. I'm sure there are lingering bugs (isn't there always?). I know this a bit much given this retro game is on the experimental side, I took some liberties because I think is a really cool and unique feature for a client. 
Regardless users should be able to "revert" to using "Webview" player (EmulatorJS w/without cores) if the standalone native core "player" implementation doesn't work properly.  Moreover, my primary goal was to keep this stuff isolated from the primary functions of Moonfin - I HOPE I at least succeeded in that.

## Type of Change
- [X] Bug fix
- [X] New feature
- [X] Refactor
- [X] Performance improvement
- [X] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Platform
- [X] Android
- [X] iOS (kinda)
- [X] tvOS (kinda)
- [X] Web
- [X] macOS (kinda)
- [X] Windows
- [ ] Linux
- [X] All / Shared code

## Testing
Mega testing via a automated harness I wrote. Plus many hours of manual testing. 

image

image

image

image

image

image

## Checklist


WizardOfYendor1:
I'm pretty much done tinkering this with PR :-).  I know it's huge, I didn't plan for it to go that way :-). 

I've done extensive testing on multiple androidtv devices, with multiple users testing for me as beta testers and it's running quite well. Lots of bugs/issues and enhancements came from that. 

I've also built, installed and tested the windows/linux clients and they works as expected. The new native core controller features aren't in there (not sure they need to be). They are only in androidtv for now but the clients function as expected. Artwork loading works great,  games run in emulatorjs and native etc... without any known issues. 

I should note N64 games are basically unplayable on the androidtv devices I tested (shield/onn  4k). They can't run in native core but do "run" in emulatorJS.... VERY slowly.  They "play" in emulatorjs but not really. Impeller may help with this but probably will create a bunch more problems not related to playing retro games.

There's a myriad of fixes and enhancements in the libretro implementation that benefits all clients regardless.

I have a local worktree that implements multiple controllers in androidtv that I am not putting into this PR, it's big enough already :-), it has dependency on all of these commits.  If/when this gets merged, I'll PR in those changes. That testing/development is near the end also, it works quite well with multi-player controllers.

RadicalMuffinMan:
First off, thank you for this. The isolation goal you set yourself was the right one and you hit it. I checked the branch against the 51 commits main has taken since you forked, and only four files overlap: app.dart, main.dart,
app_module.dart and the arb. For 149 files that's a genuinely good result. I also spot-checked your note about dropping the per-key gamepads_android call and confirmed both claims, GamepadNavigationScope.isSupported and
_readsGamepadsInDart both exclude Android. That kind of "verified before deleting" comment makes review much easier, so thanks.

I do have to stop on one thing before the code, and it isn't a code problem.

The FBNeo licence you've added in assets/licenses/cores/fbneo.txt says:

  You may not sell, lease, rent or otherwise seek to gain monetary profit
  from FB Neo

  You may not ask for donations to support your work on any project that
  uses the FB Neo source code

Moonfin has a BuyMeACoffee badge on line 14 of the README. I'm not a lawyer and I'm not going to pretend the second clause has an obvious reading when we'd be shipping a build-time-fetched binary rather than the source, but it's close enough that I'm not comfortable merging it without settling the question. None of the cores we already ship carry anything like it, so FBNeo is the outlier here rather than this being a problem with the feature.

There's also a practical angle: by your own comment in fetch_cores.sh it's ~53 MB unpacked against ~1 MB for the next biggest core, which is a lot to ask of the iOS and tvOS bundles.

A few ways forward, and I'd like your read since you've spent far more time in this than I have:

1. Merge everything except FBNeo and let arcade keep going through MAME in EmulatorJS, which this PR already supports as the fallback. That loses the native arcade path you built, which I know was the whole starting point, but the rest of the work stands on its own.
2. Ask the FBNeo maintainers directly whether distributing the core binary in a donation-funded open source client is within their intent. If they're fine with it in writing, that settles it.
3. Keep the native FBNeo path but gate it behind the user supplying their own core, so we're not the ones distributing it.

Option 1 is where I'd lean today, purely because it unblocks the other 90% of this while the license question takes as long as it takes.

Separately, and much smaller: 8055bfb6, the Watch Next foreground change, is 17 lines and affects every Android user whether they touch games or not. That one would merge on its own today and get a proper look in the process. Worth
pulling it out into its own PR regardless of how the FBNeo question lands.

I'll hold off on the deeper review of the native layer and the input dispatch until we know whether FBNeo is staying, since that changes what's in scope.

WizardOfYendor1:
@RadicalMuffinMan On the primary concern on the licensing... I had forgotten you had that button. 

I'm not a lawyer either so I don't know if this will provide any "comfort" but I would think this would be exempt for a few reasons:

1) There's no FBNeo source code or derivative works from it. We don't build it, we don't modify it.

2) We don't package/ship the FBNeo core (or any of the others). The user must do something to download it from the official repo. 

3) Not sure a voluntary "buy me a coffee"-like donations would be commercial. It's not required you pay/donate to get any of this functionality (aka exchange for goods and services). 

4)  The FBNeo license note is "You may not ask for donations to support your work on any project that uses the FB Neo **source code.**" (emphasis added) . We don't use the source code for anything, we use the official compiled binary that is user downloaded.  If we forked the source repo and created "moon-neo" and changed it etc... that would be an issue (rightfully).  

5) this is a "front end" use case, and is rather common. This is addressed by libretro (see below).

All moonfin is doing here is providing a weak front end to using the core that the user installs deliberately from the official repo. Moonfin uses the libretro's API layer to "use" the cores.  
I may have this all wrong but this is the same pattern of  retroarch.... except this one is lighter. 

Oddly enough, I was perusing this the other day researching how to best deal with analog stick snapping for certain games (analog 4-way and 8-way only games that don't expect in-between diagonal values, i.e. be dpad like - I digress...) - here's libretro's take on all  this licensing of this related to FBNeo: 

https://docs.libretro.com/library/fbneo/#license-and-changelog

This is of particular note: 
_"Actually, alternative commercial libretro frontends already exist, and we consider we are not concerned as long as they neither redistribute FBNeo nor use it as some mean of advertisement. In this scenario, only a manual installation of the core by the user will be considered legal and supported."_

My reading of that, would be you can sell frontends but cannot include FBNeo with it? i.e. the user has to install it?

Furthermore, moonfin doesn't even advertise itself as a "retro gaming" platform.... let's alone "advertising" FBNeo.  

I guess it's possible that flipping a switch to download it wouldn't be considered "manual installation".  I suppose we could remove the user initiated manual download, then make the user download it to their device(s) by some other means... and the FBNeo toggle would merely to tell moonfin that it's been installed and is available. Or perhaps provide a "upload" option like emulatorJS (i.e. user downloads it, then uploads/provides it to moonfin).
A REAL pain for sure, and I don't see how providing the user a convenient means to manually download/install a thing from the official repo *would* be considered (re)distribution (in the licensing sense). If that were true - all the scripts, package managers (apt et al), http links, etc  would all be violating that :-).


Yes teasing that all apart could be done I suppose. The problem with emulatorJS is it runs terribly for alot of games on androidTV. I'm running on a massively debloated nvidia shield and emulatorJS struggles alot.  On windows (client or brower), no problemo. In fact I've spent alot of time just making it as lean as possible and using ADB to watching device resources, threads, memory, etc, and optimizing it. Then conversely;  purposely constraining those resource and "seeing what happens" and trying to mitigate the "bad" (ANRs etc).


There's alot of good stuff in there that benefits all the cores, not just FBNeo. There's even more in a derivative branch I have that supports multi-controllers, different controller types (paddles etc) and analog support.  If this is all a problem, my inclination would be to simply remove the FBNeo download toggle/enable and hard code it disabled. Then arcade games would only have the option of running via MAME/emulatorJS. A more cautious approach would be to feature flag it so there's a convenient build "kill switch" to disable FBNeo is available should issues arise in the future.


My/our "vision" (not to speak for you) is have retro games convenient. You can watch a movie, jump over to games, goof around come back in one app/place. I could install retroarch on my devices and cut moonfin out of that path but... it's just not as convenient/cool :-). I hope there's a way to figure this out - I have to say, it's pretty cool and is working well in real world testing (thus far).

PR #744, opened 2026-07-08
fix(livetv): rebase server-owned LiveStreamFiles path to the connected base URL

brendongl:
Fixes #743.

### Problem
Live TV channels that open a managed **direct-stream** black-screen when the server is behind a reverse proxy. The server returns `MediaSource.Path` as an absolute URL built with its own **loopback bind address** (`http://127.0.0.1:8096/LiveTv/LiveStreamFiles/<id>/stream.ts`, confirmed in the Emby server log at stream-open, `IsRemote:false`). The Live TV direct-play/-stream branch used this path **verbatim**, so media3 dialed the client's own loopback:

```
java.net.ConnectException: Failed to connect to /127.0.0.1:8096
```

VOD and direct-play to genuine upstreams are unaffected. The official Emby client works because it rewrites the host to the connected server — this PR does the same.

### Fix
New pure helper `MediaStreamResolver.rebaseLiveServerPath(path, baseUrl)` in `packages/playback_core`:
- Swaps the authority (scheme+host+port) to the connected `baseUrl`, preserving path + query.
- Applies **only** when the path is server-owned: a loopback host (`127.*` / `localhost` / `::1`) **or** a `/LiveTv/LiveStreamFiles/` remux endpoint (covers container-private hosts too).
- Genuine remote upstream URLs (real HDHR/provider streams) are returned **unchanged**, preserving existing direct-play-to-upstream behavior.

Wired into both resolvers before the direct/stream classification:
- `packages/playback_emby/lib/src/emby_media_stream_resolver.dart`
- `packages/playback_jellyfin/lib/src/jellyfin_media_stream_resolver.dart`

### Tests
`test/playback/live_upstream_directplay_test.dart` — added 5 cases (loopback rebase, container-private host + query preservation, generic loopback, genuine-upstream-untouched, idempotent-when-already-published).

```
flutter test test/playback/live_upstream_directplay_test.dart
00:00 +9: All tests passed!
```

### Notes
Server-side is not cleanly fixable: Emby bakes the loopback host into the `LiveStreamFiles` path at stream-open, independent of the request Host / `X-Forwarded-*` headers, and Emby exposes no published-URI-per-subnet setting. The host rewrite on the client is the robust, server-agnostic fix (matching the official Emby client).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1134, opened 2026-08-12
Keep a working H264 profile when the Android codec probe comes back empty

brendongl:
# Pull Request

## Summary

When the Android MediaCodec probe loses its cold-start race, `_codecCapsLookDegenerate` correctly spots the bad result — but the degenerate map is applied to `PlatformDetection` anyway, before the retry is scheduled. If both retries also miss, `supportsAvc: false` / `avcMainLevel: 0` stay in place for the whole app session.

That is worse than "needless transcodes", which is what the existing comment anticipates. With `supportsAvc` false, the `if (supportsAvc && avcMainLevel > 0)` block in `device_profile_builder.dart` never runs, so the device profile carries **no h264 codec profiles at all**. The server therefore receives:

```
&h264-profile=none
```

and can match **no encoder — hardware or software**:

```
>>>>>>  FindVideoEncoder - MediaType: h264, UseHardwareCodecs: True, HWA-Mode: Automatic
Info    Checking: 'QuickSync DG2 Arc A310 [1] - H.264 (AVC)'
NoMatch Encoder does not support any of the requested client profiles ()
NoMatch Encoder does not match
...
Info    Checking: 'x264'
NoMatch Encoder does not support any of the requested client profiles ()
NoMatch Encoder does not match

FfRunException: No video encoder found for 'h264'
```

Every `/videos/{id}/hls1/main/N.ts` returns **HTTP 500**, playback dies, the app restarts the stream, direct-plays until it next needs a transcode, and 500s again. To the user it reads as buffering on a several-minute rhythm rather than as an error.

This PR keeps a usable H264 profile in place whenever the probe comes back empty, so the worst case degrades to an unnecessary transcode instead of an unplayable stream.

### How it was found

An Emby deployment where one customer reported "buffering every 3–4 minutes". Server-side metrics looked healthy throughout — only 4 concurrent transcodes at the time of the failures, so this was not encoder saturation.

Over a 2-day server log window:

| `h264-profile=` value | requests |
|---|---:|
| `high,main,baseline,constrainedbaseline` | 2254 |
| `high,main,baseline,constrainedbaseline,high10` | 180 |
| **`none`** | **32** |

All 32 came from two clients, and the affected user's bad-segment rate was 9.7% against a fleet median under 1%. Decisively: the **same physical box was fine on the stock Emby Android app at the same moment from the same IP** — only Moonfin sent `none`, which places it in the client's device profile rather than in the server or the network.

## Related Issues

No existing issue — found from a production incident and filed directly as a PR.

- Closes #
- Fixes #
- Related to #

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

- **Never apply degenerate caps as-is** (`lib/main.dart`). New `_withAvcFloor` fills in `supportsAvc: true` and `avcMainLevel: 41` (Level 4.1 / 1080p) and leaves every other probed field untouched. 4.1 is the floor any Android device running this app clears, so the repair cannot claim decoding the device lacks.
- **Retry harder** (`lib/main.dart`). 4 attempts with exponential backoff (2s/4s/8s/16s) rather than 2 fixed 2s attempts — enumeration loses the race precisely when the device is busiest after boot. A probe that throws no longer aborts the remaining attempts, since one failure says nothing about the next.
- **Cover the legacy Dolby-Vision fallback path** (`lib/main.dart`). It only ever reports DoVi, so reaching it leaves the profile with no h264 support — the identical broken state.
- **Expose the capability map for whole-map inspection** (`lib/util/platform_detection.dart`). New `PlatformDetection.mediaCodecCapabilities` getter returns an unmodifiable view, so the repair can read and re-apply the map rather than probing key by key.
- **Tests** (`test/playback/device_profile_builder_test.dart`). Two cases at the layer where the bug is observable.

All retries stay off the launch path, and the device profile is built per playback, so a corrected result still fixes the next playback without a restart.

## Platform
- [x] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] All / Shared code

The affected code is the Android MediaCodec probe in `main.dart`; no other platform reaches it.

## Testing

Two cases added to `test/playback/device_profile_builder_test.dart`, at the layer where the bug is observable:

- an AVC device advertises the four `VideoProfile` values the server matches an encoder against;
- caps reporting no AVC advertise none — pinning the exact state that produces `h264-profile=none`, so a future change that reintroduces it fails the suite.

`flutter test test/playback/device_profile_builder_test.dart` passes (56/56). Full-suite failures are unrelated and reproduce identically on an unmodified checkout of this branch point (`plugin_sync_service_*`, `offline_items_api`).

- [x] Not tested (explain why): the trigger is a cold-start enumeration race that cannot be forced on demand, so the fix is covered by unit tests at the layer where the bug is observable rather than by a device repro. The *bug* itself is evidenced from production server logs (the `h264-profile=none` counts above); on-device confirmation of the fix has to wait for a build to reach the affected hardware.

### Test Steps
1. `flutter test test/playback/device_profile_builder_test.dart` — 56/56 pass.
2. To verify on hardware once built: on an Android device that has reproduced the degenerate cold-start probe, launch the app and start a title that requires a transcode; watch the server access log for the `h264-profile=` parameter.
3. Before the change the parameter reads `h264-profile=none` and the `hls1/main/N.ts` segments return HTTP 500; after it, the profile list is populated, an encoder is matched, and the segments return 200.

N/A — no UI change.

## Checklist

## Note on authorship

This patch was written by an AI agent (Claude Code), from a real production incident, and reviewed by the repository owner before submission. Please review it with that in mind — particularly the choice of Level 4.1 as the fallback floor, which is a judgement call rather than a measured value, and would be easy to argue up or down.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -32,6 +32,7 @@ import 'playback/appletv_backend.dart';
 import 'playback/audio_capability_profile.dart';
 import 'playback/audio_capability_probe.dart';
 import 'playback/audio_handler.dart';
+import 'playback/codec_caps_repair.dart';
 import 'playback/media_browse_service.dart';
 import 'playback/mpris_service.dart';
 import 'playback/playback_lifecycle_handler.dart';
@@ -188,15 +189,6 @@ Future<void> _detectAndSetDisplayCapabilities() async {
   } catch (_) {}
 }
 
-/// A cold-start probe can race codec enumeration and return a map with no
-/// usable H264 support. Every Android device that reaches this code plays
-/// H264, so such a result is a transient failure, not a real capability.
-bool _codecCapsLookDegenerate(Map<String, dynamic> caps) {
-  final supportsAvc = caps['supportsAvc'] == true;
-  final avcMainLevel = caps['avcMainLevel'];
-  return !supportsAvc || avcMainLevel is! int || avcMainLevel <= 0;
-}
-
 Future<Map<String, dynamic>?> _queryCodecCaps(MethodChannel channel) async {
   final raw = await channel.invokeMethod<Map<dynamic, dynamic>>(
     'mediaCodecCapabilities',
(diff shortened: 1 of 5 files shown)


jmawet:
Please use the PR template

RadicalMuffinMan:
@brendongl just like I told you last time, [have your AI use the template](https://github.com/Moonfin-Client/Moonfin-Core/pull/744#issuecomment-4921380539). Once done, I'll re-open the PR and review it.

brendongl:
Description rewritten to use `.github/pull_request_template.md` — all sections filled in (Summary, Related Issues, Type of Change, Changes Made, Platform, Testing + Test Steps, Screenshots, Checklist). Sorry for the repeat.

I can't reopen it from my side (GitHub rejects the reopen for a PR closed by a maintainer), so it's over to you when you have a moment. The branch is unchanged at `84465634` and still merges cleanly against `main`.

RadicalMuffinMan:
Thanks AI bot

PR #1144, opened 2026-08-13
add address redaction for diagnostic logs

broken-droid:
# Pull Request

## Summary
Some people are sensitive about their server urls, this strips them out if they need to send a log.
ex: 
2026-08-12T19:15:05.905684 DEBUG [network] ← 200 GET http://[REDACTED]/Moonfin/Games/Libraries
2026-08-12T19:15:05.924406 DEBUG [media] Media3: video size 0x0
2026-08-12T19:15:06.949493 DEBUG [network] ← 204 POST http://[REDACTED]/Sessions/Playing/Stopped
2026-08-12T19:15:14.320364 INFO  [general] Uploading diagnostic report (67 entries) to http://[REDACTED]

## Related Issues

- Closes #
- Fixes #
- Related to #

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
-added _redactRegex - static regex to match anything between http(s?):// and the next / or end of string
-added _redact function - uses the regex to replace addresses in a string
-modified log function to use _redact on message and error strings

## Platform
- [ ] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] All / Shared code

## Testing

### Test Steps
1. Turn on diagnostic logging
2. Do stuff
3. Upload
4. Verify addresses are redacted

## Checklist

### DIFF
--- a/lib/data/services/log_service.dart
+++ b/lib/data/services/log_service.dart
@@ -72,7 +72,10 @@ class LogService extends ChangeNotifier {
 
   static const int _maxEntries = 2000;
 
-  static final _redactRegex = RegExp(r'(https?://)[^/\s]*(?=/|$)');
+  static final _redactRegex = RegExp(
+    r'((?:https?|wss?)://)[A-Za-z0-9._~%\-:@\[\]]+',
+    caseSensitive: false,
+  );
 
   final UserPreferences _prefs;
   final MediaServerClientFactory _clientFactory;
@@ -255,7 +258,7 @@ class LogService extends ChangeNotifier {
   };
 
   String _redact(String text) {
-    if (text.isEmpty || !text.contains('://')) {
+    if (!text.contains('://')) {
       return text;
     }
     return text.replaceAllMapped(_redactRegex, (match) {
--- a/lib/data/services/log_service.dart
+++ b/lib/data/services/log_service.dart
@@ -72,6 +72,8 @@ class LogService extends ChangeNotifier {
 
   static const int _maxEntries = 2000;
 
+  static final _redactRegex = RegExp(r'(https?://)[^/\s]*(?=/|$)');
+
   final UserPreferences _prefs;
   final MediaServerClientFactory _clientFactory;
   final DeviceInfo _deviceInfo;
@@ -143,8 +145,8 @@ class LogService extends ChangeNotifier {
       time: DateTime.now(),
       level: level,
       category: category,
-      message: message,
-      error: error?.toString(),
+      message: _redact(message),
+      error: error != null ? _redact(error.toString()) : null,
     );
     _entries.addLast(entry);
     while (_entries.length > _maxEntries) {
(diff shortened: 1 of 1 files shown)


lib/data/services/log_service.dart
```diff
@@ -72,6 +72,8 @@ class LogService extends ChangeNotifier {
 
   static const int _maxEntries = 2000;
 
+  static final _redactRegex = RegExp(r'(https?://)[^/\s]*(?=/|$)');
```
RadicalMuffinMan: I like this, its a pretty great idea but the issue might be the (?=/|$) at the end. After it matches the host it peeks at the next character and only accepts a slash or the end of the line. That works when the URL has a path on it, because the next character is the slash but when the URL ends at the host and the log message carries on around it, the next character is whatever the sentence had next, and the match fails. Nothing gets redacted at all in that case. So these still leak:

```
GET http://192.168.1.5:8096 (connectionTimeout)
probe http://myserver.duckdns.org:8920, falling back
```
Also https? misses ws:// and wss://. The socket code builds ws://host/socket?ApiKey=... and when it fails to connect that whole thing ends up in the log.

```suggestion
    static final _redactRegex = RegExp(
      r'(https?|wss?)://[A-Za-z0-9._~%\-:@\[\]]*',
      caseSensitive: false,
    );
```

lib/data/services/log_service.dart
```diff
@@ -251,4 +253,13 @@ class LogService extends ChangeNotifier {
...
   };
+
+  String _redact(String text) {
+    if (text.isEmpty || !text.contains('://')) {
+      return text;
+    }
+    return text.replaceAllMapped(_redactRegex, (match) {
+      return '${match.group(1)}[REDACTED]';
+    });
+  }
```
RadicalMuffinMan, line 267: Small follow on from the regex change. First one is now just the scheme without the ://, so that moves into the replacement or you end up with http[REDACTED].

Also dropped the isEmpty check since contains already handles an empty string.

```suggestion
  String _redact(String text) {
    if (!text.contains('://')) {
      return text;
    }
    return text.replaceAllMapped(_redactRegex, (match) {
      return '${match.group(1)}://[REDACTED]';
    });
  }
```

RadicalMuffinMan (changes requested):

lib/data/services/log_service.dart
broken-droid: changed it to also capture the ://, and + instead of * at the end

lib/data/services/log_service.dart
broken-droid: removed the isEmpty check, and capture includes the separator now

PR #1166, opened 2026-08-16
fix(androidtv): stop stale video sessions after backgrounding

bricenice17:
# Pull Request

## Summary

Fixes stale Jellyfin playback sessions when Moonfin remains backgrounded on Android TV, without discarding local VOD resume state.

After the existing three-second background grace period, Moonfin now ends the obsolete server playback claim and releases the current backend/source while retaining the queue item and saved position needed for a safe foreground return. Movies and episodes fresh-resolve at their saved position; Live TV fresh-resolves at the live edge.

This also isolates progress reporting by playback generation so a late request from the retired session cannot revive or overwrite the replacement session.

Audio background playback, offline playback, non-TV platforms, quick background/resume, and the canonical user stop behavior remain unchanged.

## Related Issues

No existing issue found.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

- Added a narrow Android TV background-stop path that ends stale server ownership without clearing the queue or stored VOD position.
- Preserved the existing three-second grace period and cancellation on quick foreground return.
- Added item-ownership checks so a stale timer cannot stop or restart a replacement queue item.
- Fresh-resolve movies and episodes at the saved position after timeout; fresh-resolve Live TV at the live edge.
- Track progress requests by playback generation and compensate late completions with a stop scoped to the retired session.
- Keep cleanup bounded even when a progress request never completes.
- Preserve the existing destructive canonical stop behavior for explicit/user stop flows.
- Added lifecycle, resume, replacement-item, overlapping-progress, stalled-request, and canonical-stop regression tests.

## Platform
- [x] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] All / Shared code

## Testing

### Automated checks

- `flutter test test/playback` — 252 tests passed
- Focused lifecycle/stale-session rerun — 23 tests passed
- Touched-file `flutter analyze --no-fatal-infos` — no new findings; four pre-existing info-level lints remain in `playback_manager.dart`
- `git diff --check`
- GitHub Android build — passed (28m30s)
- Independent implementation review focused on stale-session races and bounded cleanup

### Physical validation

Tested on the side-by-side Moonfin beta on an Amazon AFTKA Fire TV against Jellyfin 10.11.11:

1. Played a movie, episode, and Live TV.
2. Exited/backgrounded Moonfin and also powered off the TV without explicitly exiting playback.
3. Confirmed each obsolete Jellyfin session was removed after the grace period.
4. Confirmed movie and episode playback retained the previous resume position.
5. Confirmed Live TV returned through a fresh session at the live edge.
6. Confirmed quick foreground return still cancels the pending cleanup.

Before this fix, Live TV sessions could remain active indefinitely after power-off. With this change, stale server ownership ends while resumable local VOD state survives.

## Screenshots

Not applicable. This is a playback lifecycle change with no UI changes.

## Checklist

### DIFF
--- a/lib/playback/playback_lifecycle_handler.dart
+++ b/lib/playback/playback_lifecycle_handler.dart
@@ -12,6 +12,7 @@ class PlaybackLifecycleHandler with WidgetsBindingObserver {
   // right after the app goes to the background. _restoreState runs the same
   // correction on resume, so ten seconds of ticks is enough.
   static const int _maxBgCorrectionTicks = 40;
+  static const Duration _videoBackgroundGracePeriod = Duration(seconds: 3);
 
   final PlaybackManager _manager;
   Duration? _savedPosition;
@@ -20,6 +21,8 @@ class PlaybackLifecycleHandler with WidgetsBindingObserver {
   int _bgCorrectionTicks = 0;
   bool _bgSeekInFlight = false;
   bool _screenLocked = false;
+  Timer? _videoStopTimer;
+  AggregatedItem? _videoStopItem;
 
   void setScreenLocked(bool locked) {
     _screenLocked = locked;
@@ -31,6 +34,10 @@ class PlaybackLifecycleHandler with WidgetsBindingObserver {
 
   @override
   void didChangeAppLifecycleState(AppLifecycleState state) {
+    if (state == AppLifecycleState.resumed) {
+      _cancelVideoStop();
+    }
+
     final currentItem = _manager.queueService.currentItem;
     bool isAudio = false;
     if (currentItem is AggregatedItem) {
@@ -40,7 +47,8 @@ class PlaybackLifecycleHandler with WidgetsBindingObserver {
       if (meta != null) {
         final type = meta['Type']?.toString();
         final mediaType = meta['MediaType']?.toString();
-        isAudio = type == 'Audio' || type == 'AudioBook' || mediaType == 'Audio';
+        isAudio =
+            type == 'Audio' || type == 'AudioBook' || mediaType == 'Audio';
       }
     }
     if (isAudio) {
@@ -57,6 +65,7 @@ class PlaybackLifecycleHandler with WidgetsBindingObserver {
         if (backend is Media3PlayerBackend) {
           unawaited(backend.appPaused());
         }
+        _scheduleVideoStop(currentItem);
         break;
       case AppLifecycleState.resumed:
         final backend = _manager.backend;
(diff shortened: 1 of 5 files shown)


RadicalMuffinMan:
Thanks for this and especially for the testing. Trying it on real hardware against a real server, and checking the upstream session actually cleaned itself up afterwards, is more than most playback changes get.

The scoping is careful too. Audio, offline and every other platform are excluded properly, and you re-check the item still owns the timer before stopping it. I also went looking for the obvious trap, picture-in-picture
getting stopped while the user is still watching it, and it can't happen here since PiP is phone-only and this is TV-only.

One thing I'd like to sort out before merging.

The stop you're calling doesn't just end the session, it also clears the queue and resets playback state. So if someone comes back after four seconds, their place is gone. Right now backgrounding and returning just picks up where they
were.

Three seconds is short for a TV. The Assistant overlay is one mic press away on most remotes, and system dialogs, input switches and a mispressed Home all background us while the user is still sitting right there. Any of those
running a moment long now costs them their spot in the episode.

The problem you're actually fixing is that we keep telling Jellyfin we're playing. So the narrower fix would be to end the server session and stop reporting progress, but leave local playback alone. Then coming back late is
harmless, and the exact timing matters a lot less.

If you'd rather keep the full stop, that's fair, but then I'd want the grace period much longer, more like 30 to 60 seconds. Three only makes sense if being wrong is cheap, and here it isn't.

Either way it's a change to what the timer does at the end, not to the structure you've built, which is solid. Let me know which way you'd rather take it.

Two small things while you're in there: the item reference isn't cleared when the timer fires, only on resume, which is harmless but worth a comment saying so and a couple of the reformatted lines are our formatter version mismatch
rather than your change, so reverting those would keep the diff to just the fix.

bricenice17:
@RadicalMuffinMan Yea the narrower fix is the better approach. I’ll change the timeout path so it ends the stale Jellyfin playback session and stops progress reporting without clearing the queue or losing the current item, position, or play/pause intent. When Moonfin resumes, it will fresh-resolve playback and continue the same item from the saved position instead of treating the timeout as a user stop.

With that change, 3 seconds is no longer the difference between preserving playback and losing the user’s place. Crossing it may require a fresh resolution and brief rebuffer on return, but it won’t cost the user their place. I can keep the existing 3 seconds so abandoned sessions and server resources are cleaned up promptly while still filtering short lifecycle transitions. BUt if you want it longer, whatever you're cool with is fine by me. 

I’ll also add a comment explaining that the item reference is intentionally retained after the timer fires until resume, since it's used to verify ownership before restoring playback. And I’ll revert the formatter only changes so the diff stays focused. I’ll update the tests and validate the revised behavior on the Android TV beta before pushing the PR update.

PR #1196, opened 2026-08-19
feat(seerr): add duotone genre backdrops for seerr rows

### LINKED ISSUE #1195: [Feature] Seerr genres rows duotone backdrops

The Seerr genre rows repeat the same handful of films, so the cards are hard to tell apart. For example Animation, Comedy and Family could have the same backdrop.

Image

Seerr's own rows don't have this problem: each genre has its own colour, so the cards stay distinct even when two land on the same film.

### Proposed solution

Build the backdrop the way Seerr's client does: read `backdrops[4]` instead of the first entry, since the list comes back sorted by popularity and a popular film sits in several genres, and pass the path through TMDB's duotone filter with a colour pair per genre.

There is also a film/show rows matching bug: `_cachedRowImageUrl()` in `home_screen.dart` remembers the image URL it worked out for an item, filed under `serverId|id|…`. A genre card's `item.id` is the TMDB genre id, which is the same in both rows, so the show row gets handed the film row's image. Adding the media type to that name fixes it.

Also print the genre name across the middle of the artwork, like Seerr and like the Jellyfin genre row.

Proposed PR: #1196 

### Platform

All / Not platform-specific

### Alternatives considered

_No response_

### Additional context

_No response_

bientavu:
# Pull Request

## Summary
Seerr genre cards all looked alike. Genres that share a popular film got the same backdrop, because we took `backdrops.first` and the most popular film is exactly the one several genres have in common. On top of that, the film row and the show row showed identical images: `home_screen.dart` remembers image URLs under the item's id, and for a genre card that id is the TMDB genre id, the same in both rows. Cards now build their backdrop the way Seerr's own client does, through TMDB's duotone filter, so every genre has its own colour and stays distinct even when two land on the same film.

## Related Issues

- Closes #1195 
- Fixes #
- Related to #

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [x] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
- New `lib/util/seerr_genre_art.dart`: `seerrGenreBackdropUrl()` builds the TMDB `w1280_filter(duotone,…)` URL for a genre. The colours and the genre they belong to are taken from Seerr, so a genre looks the same here as it does there. It reads `backdrops[4]`, again matching Seerr: the first few entries are the most popular films, and those are what repeat across genres.
- New `lib/ui/widgets/seerr/seerr_genre_label.dart`: the genre name across the middle of the card, with the backdrop darkened 30% behind it, since some duotone backdrops come out light enough that white text is hard to read on them. It uses the same lettering as the Jellyfin genre row so the two rows look like they belong to the same app.
- Add `MediaType` to the name `_cachedRowImageUrl()` files its results under. Without it the show row gets handed the film row's image for every genre the two have in common.

## Platform
- [ ] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [x] All / Shared code

## Testing

### Test Steps
1. Enable both Seerr genre rows on Home. Each card has its own colour, and the repeats between genres are gone.
2. The genre name reads across the middle of the artwork and also under the card, matching the Jellyfin genre row.
3. Seerr --> Discover: the genre cards match the home rows. The shortcut cards next to them are unchanged.

### Before
Screenshot 2026-08-19 at 18 14 54

### After (two screenshots as my branch include the #1184 change)
Classic:
Screenshot 2026-08-19 at 17 31 26
Modern:
Screenshot 2026-08-19 at 17 32 04

## Checklist

### DIFF
--- a/lib/ui/screens/home/home_screen.dart
+++ b/lib/ui/screens/home/home_screen.dart
@@ -72,6 +72,7 @@ import '../../navigation/route_lifecycle_observer.dart';
 import '../../util/home_row_title_localizer.dart';
 import '../../../util/game_library.dart';
 import 'home_view_model.dart';
+import '../../widgets/seerr/seerr_genre_label.dart';
 
 Color get _homeBackground => AppColorScheme.background;
 
@@ -3607,8 +3608,12 @@ class _ContentRowsState extends State<_ContentRows>
     double requestScale, {
     bool isMyMediaRow = false,
   }) {
+    // MediaType belongs in the key: a Seerr genre card takes the TMDB genre id
+    // as its item id, and a genre in both the movie row and the series row has
+    // the same id in each, so the second row would reuse the first row's image.
     final key =
-        '${item.serverId}|${item.id}|${imageType.index}|${height.round()}'
+        '${item.serverId}|${item.id}|${item.rawData['MediaType']}'
+        '|${imageType.index}|${height.round()}'
         '|$useSeriesThumbs|${requestScale.toStringAsFixed(2)}|$isMyMediaRow';
     final cached = _rowImageUrlCache[key];
     if (cached != null || _rowImageUrlCache.containsKey(key)) {
@@ -4746,8 +4751,15 @@ class _ContentRowsState extends State<_ContentRows>
             cardSubtitleWidget = null;
           }
 
+                  // Seerr genre cards print their name across the artwork,
+                  // the way the Jellyfin genre row does.
+                  final isSeerrGenreCard =
+                      _isSeerrFilterRow(row) && item.type == 'Genre';
                   final card = MediaCard(
                     title: cardTitle,
+                    imageOverlays: isSeerrGenreCard
+                        ? [Positioned.fill(child: SeerrGenreLabel(name: item.name))]
+                        : const <Widget>[],
                     subtitle: cardSubtitle,
                     subtitleWidget: cardSubtitleWidget,
                     imageUrl: imageUrl,
--- a/lib/ui/screens/home/home_view_model.dart
+++ b/lib/ui/screens/home/home_view_model.dart
@@ -36,6 +36,7 @@ import '../../../preference/seerr_preferences.dart';
 import '../../../data/viewmodels/seerr_discover_view_model.dart';
 import '../../widgets/seerr/seerr_shortcuts.dart';
 import '../../../data/services/custom_external_lists_service.dart';
+import '../../../util/seerr_genre_art.dart';
 
 class HomeViewModel extends ChangeNotifier {
   /// Cards in a merged row, matching what a single library row shows.
@@ -2249,7 +2250,7 @@ class HomeViewModel extends ChangeNotifier {
               _seerrFilterItem(
                 id: g.id.toString(),
                 name: g.name,
-                imagePath: g.backdrops.isNotEmpty ? g.backdrops.first : '',
+                imagePath: seerrGenreBackdropUrl(g.id, g.backdrops) ?? '',
                 type: 'Genre',
                 mediaType: 'movie',
                 filterType: 'genre',
@@ -2264,7 +2265,7 @@ class HomeViewModel extends ChangeNotifier {
               _seerrFilterItem(
                 id: g.id.toString(),
                 name: g.name,
-                imagePath: g.backdrops.isNotEmpty ? g.backdrops.first : '',
+                imagePath: seerrGenreBackdropUrl(g.id, g.backdrops) ?? '',
                 type: 'Genre',
                 mediaType: 'tv',
                 filterType: 'genre',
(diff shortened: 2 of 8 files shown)


mattsigal:
I think these look great! One small thing I noticed though -- it looks like the text size is based on Classic Rows (where it looks great), but from your screenshots, it looks quite tiny on Modern?

2026-08-19_11-47-40_brave

Can you tweak the modern rows to have a more proportionally sized font?


EDIT: For comparison, this is from Seerr itself:
2026-08-19_11-51-56_brave

bientavu:
Hi @mattsigal yep I agree, it does look a bit tiny, it's because I took the same sizing as the regular Genres row: https://github.com/bientavu/Moonfin-Core/blob/feat/seerr-genres-backdrops/lib/ui/widgets/media_card.dart#L706

Screenshot 2026-08-19 at 21 39 36

Sure, I can tweak that to make it more proportional, and also apply it to the default Genres row if that's ok? Or would you rather keep it as it is? 🙂

mattsigal:
> Hi @mattsigal yep I agree, it does look a bit tiny, it's because I took the same sizing as the regular Genres row: https://github.com/bientavu/Moonfin-Core/blob/feat/seerr-genres-backdrops/lib/ui/widgets/media_card.dart#L706

Thanks! I think the worry about the Moonfin Modern genre rows is that the text needs to look nice in poster *and* thumbnail (where as the Seerr rows always have the larger area). This problem is kind of exacerbated by localization -- some of the other languages have labels that are much longer than English/French, so it is a bit tricky.

Perhaps the nicest solution would be the smaller font on the posters (e.g., ~~Science-Fiction~~ Action & Adventure in your screenshot already takes up almost the entire horizontal space), but to expand it on the focus?

bientavu:
@mattsigal hmm I see what you mean. I like your suggestion, I'll do that tomorrow and show you how it looks!

bientavu:
Hi @mattsigal, I pushed the changes with your suggestion: the genre label is now larger on the thumbnails, for both the default Genres row and the Seerr genre rows. I kept it a bit restrained so it doesn't end up oversized on TV screens. We can manage the size with [this value](https://github.com/bientavu/Moonfin-Core/blob/feat/seerr-genres-backdrops/lib/ui/widgets/media_card.dart#L115)

Screenshots below, what do you think?

Screenshot 2026-08-20 at 09 54 39
Screenshot 2026-08-20 at 09 55 08

mattsigal:
Looking better, I think! Not trying to be nit-picky but I am wondering about an alternative for the library-based rows (for both Classic and Modern) for you to consider.

The Seerr rows really have great legibility in terms of their contrast due to the colored overlay. I think they look great, although the font size could probably be made a little larger to be a bit closer to Seerr:

2026-08-20_03-52-25_brave
2026-08-20_03-52-39_brave

Regardless of that, the comparison between the Seerr and the Genre row is really stark in the sense that the Genre row is quite busy, especially when we are showing posters, which have a lot of text themselves. However, the purpose of these cards is not to be links to those particular movies like they would be on Recently Added or a row like that, but to just be somewhat representative art for the genre. So I'm almost wondering if it would make sense to:

1. Add an opacity filter to all the images on the Genres row to darken the poster (similar to what we do on the Modern Details page). We don't need a slider like we have there but using a value that makes the overlaid text "pop" would be nice:
2026-08-20_03-48-18_brave

2. Then, reposition the labels to be lower to have less overlap with the "focal point" of the posters and increase the font size for the selected poster (just on Modern rows), to show something like this:
638729759-31ef5d3c-a7fc-4577-931a-2d7aa410974d - Copy

My thinking is that would differentiate the Seerr vs. non-Seerr genre rows rather nicely and increase the text legibility on the later somewhat (especially once the opacity filter is also applied). In any case, since posters are not standardized in terms of their layouts, there won't be a perfect solution here so I'm just thinking out loud :) Thanks for working on this and all the tinkering!

@RadicalMuffinMan -- any thoughts or preferences on the positioning here?

bientavu:
Hi @mattsigal, no problem your thoughts are interesting 🙂

Regarding the opacity filter for the default Genres row, [I believe it's already in place here](https://github.com/bientavu/Moonfin-Core/blob/feat/seerr-genres-backdrops/lib/ui/widgets/media_card.dart#L717). To make it easier to see, I bumped the value up to 0.8. I also moved the label below the center as you suggested. For reference, here's what that looks like:
Screenshot 2026-08-20 at 14 19 10
Screenshot 2026-08-20 at 14 17 32

What do you both think? @mattsigal @RadicalMuffinMan 
I agree it's worth making it obvious that the Genres and Seerr rows aren't the same thing. The lower label isn't an obvious win to me though, since posters put their own text in different places. Making the Genres posters a bit darker does make sense. I used 0.8 here just to show the difference clearly, but we can obviously lower it a bit.

EDIT:
Extra thought to differentiate Genres and Seerr genres rows: removing the media card title for Seerr genres rows. Screenshots below
Screenshot 2026-08-20 at 14 52 03
Screenshot 2026-08-20 at 14 51 13

jmawet:
I don't the media card title should be removed for any rows. It's strange having all rows have titles except this seerr row. 

image

Honestly I like what you did here with the text size. I think this is good enough for this PR at least. 

If we really want to tweak the library genres rows we can do that separately after more discussion internally. For now I think we should leave things as is and just merge @bientavu's changes for the color overlay and text size.

RadicalMuffinMan:
Another person who likes the extended navbar labels! Second one I see today, I'm glad people actually use that.

PR #1222, opened 2026-08-21
fix(l10n): declare Old Persian locale

wmurphy126:
# Pull Request

## Summary
Adds an explicit `@@locale` declaration to the Old Persian localization file so Flutter can correctly identify the locale during localization generation.

Current `main` fails during `flutter-tvos pub get` / `gen-l10n` with:

`The following .arb file's locale could not be determined: lib/l10n/app_peo.arb`

This PR declares the locale explicitly as `peo`.

## Related Issues
Link related issues or tickets separated by commas.

- Closes #
- Fixes #
- Related to #

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made
List the key changes included in this PR.

- Added `@@locale: peo` to `lib/l10n/app_peo.arb`.
- Allows Flutter localization generation to identify the Old Persian locale instead of failing during dependency resolution.
- No other source or workflow files are changed.

## Platform
- [ ] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [x] All / Shared code

## Testing
Describe how this change was tested.

- [ ] Tested on emulator / simulator
- [ ] Tested on physical device
- [ ] Manual testing completed
- [ ] Not tested (explain why):

A tvOS TestFlight build containing this fix is currently running. The failure was first reproduced against current `main`, where Flutter localization generation stopped on `app_peo.arb`.

### Test Steps
1. Run the tvOS build from current `main` and observe `flutter-tvos pub get` fail because the locale for `lib/l10n/app_peo.arb` cannot be determined.
2. Add the explicit `@@locale: peo` declaration.
3. Re-run the same tvOS TestFlight build and verify localization generation succeeds and the workflow proceeds past dependency resolution.

## Screenshots (if applicable)
N/A — no UI changes.

## Checklist
- [ ] Code builds successfully
- [x] Code follows project style and conventions
- [x] No unnecessary commented-out code
- [ ] No new warnings introduced

PR #1224, opened 2026-08-21
fix(l10n): restore the {type} placeholder in Finnish

DrPleb:
# Pull Request

## Summary
main does not build from source. The Finnish translation of `requestSeriesOrMovie4k` localised the placeholder name itself — the message reads `Pyydä 4K {tyyppi}` while the metadata in the same block still declares `type`. gen-l10n merges placeholders across every locale, so the generated method becomes `requestSeriesOrMovie4k(String type, Object tyyppi)` and the one-argument call in `seerr_request_dialog.dart:42` no longer compiles.

The committed `app_localizations.dart` predates the translation, so the tree looks healthy until l10n is regenerated — which `flutter pub get` does on every build.

Introduced by 75ebc836 (Translated using Weblate (Finnish)).

## Related Issues
- Related to #1222

## Type of Change
- [x] Bug fix

## Changes Made
- Restore the `{type}` placeholder in `app_fi.arb` so it matches the declared metadata
- Declare `@@locale` in `app_peo.arb`, the same one-line change as #1222, without which CI cannot reach this fix. Happy to drop that commit and rebase if #1222 lands first.

## Platform
- [x] All / Shared code

## Testing

### Test Steps
1. On main, run `flutter pub get` — l10n generation fails on `app_peo.arb`.
2. With only the peo declaration, the build then fails compiling `requestSeriesOrMovie4k`.
3. With both changes, `flutter pub get` and `flutter analyze` are clean and the app builds and runs on a physical Android device.

## Checklist

### DIFF
--- a/lib/l10n/app_fi.arb
+++ b/lib/l10n/app_fi.arb
@@ -9281,7 +9281,7 @@
     "@onWatchlist": {
         "description": "Details screen button label when the title is already on the Seerr watchlist"
     },
-    "requestSeriesOrMovie4k": "Pyydä 4K {tyyppi}",
+    "requestSeriesOrMovie4k": "Pyydä 4K {type}",
     "@requestSeriesOrMovie4k": {
         "description": "Bottom sheet title for requesting media in 4K",
         "placeholders": {
--- a/lib/l10n/app_peo.arb
+++ b/lib/l10n/app_peo.arb
@@ -1,4 +1,5 @@
 {
+    "@@locale": "peo",
     "appTitle": "Moonfin",
     "@appTitle": {
         "description": "The application title"


jmawet:
Please use the PR template

DrPleb:
Both Apple builds fail on an external dependency, unrelated to these
branches:

    could not find the commit b783ae0c3d25c634dc4f1fab3c7a563b6da11967
    in https://github.com/RadicalMuffinMan/AetherEngine

iOS and macOS resolve the same Swift package, so both stop at the same
point. The pinned revision no longer exists upstream. The other five
platforms build.

DrPleb:
> Please use the PR template

RadicalMuffinMan:
Superseded by f6186a92ab3a2853ea7401b4b5da16065808a021

Issue #1205, opened 2026-08-20
[Playback] Chromecast: HEVC sources fail even though the server offers an H.264 transcode


DrPleb:
### Moonfin version

2.4.0

### Platform

Android

### OS version

Android 14

### Device model

OPPO A78 4G

### Input method

Touch

### Server type and version

Jellyfin 10.11.11

### Content type

Movie

### Media details (as much as you know, check the info button at the bottom right of the player)

Container:  MKV
Resolution: 1920x1080
HDR:        None (SDR, 10-bit)
Video:      HEVC Main 10 @ L4.0, 23.976 fps, 13.77 Mbps
Audio:      AAC LC 2.0 stereo, 128 kbps  (stream index 5 — explicitly selected)
            file also has DTS-HD MA 7.1 (index 4, default) — ruled out, see below
Subs:       SRT eng + ita, both disabled during the test (subtitleStreamIndex=-1)
Bitrate:    ~13.8 Mbps video

### Playback path

Transcode

### Symptoms (check all that apply)

- [ ] Black screen / endless spinner
- [ ] Audio only (no video)
- [ ] Stuttering / frame drops
- [ ] A/V sync issues
- [ ] Seeking/scrubbing broken
- [ ] Subtitles missing or incorrect
- [ ] Unexpected forced transcode
- [ ] App crashes / exits

### Steps to reproduce

Receiver: Chromecast 3rd gen (H.264 High up to 1080p60, VP8 — no HEVC decoder)

1. Cast a title whose SOURCE is H.264 1080p + AAC 2ch.
   -> Works. Server transcodes to h264_qsv / mpegts, Chromecast plays it.

2. In the same app session, same Chromecast, cast a title whose SOURCE is
   HEVC 1080p. Select the AAC 2ch audio track manually.
   -> App shows "unable to cast". Playback never starts.

Both files are above the server's 4 Mbps remote cap, so the server must transcode
both, and with HEVC encoding disabled server-side it offers H.264 for both. The
app receives HTTP 200 from PlaybackInfo and then makes no stream request at all —
ffmpeg is never invoked. It appears to decide from the source file's codec rather
than from the MediaSource the server returned.

Variables ruled out:
- Audio: retested with the AAC 2ch track instead of the default DTS 8ch. Same failure.
- Resolution: the failing title is 1080p, same as the working one.
- Container/subs: both MKV, subtitles disabled in both.
The only remaining difference is the source video codec.

[...]

PR #1225, opened 2026-08-21
fix(cast): cap H.264 at level 4.1 for Cast receivers

DrPleb:
# Pull Request

## Summary
`chromecastDeviceProfile()` allowed H.264 up to level 4.2. Cast generations 1 through 3 stop at High profile level 4.1 — only Ultra and Google TV go past it — so the receiver rejects the HLS manifest outright.

Casting a 1080p HEVC source to a 3rd gen Chromecast, the master playlist advertised:

    CODECS="avc1.64002A"   High, level 4.2   -> rejected
    CODECS="avc1.640029"   High, level 4.1   -> plays

The rejection happens before a single segment is requested, so the server never starts a transcode at all: no ffmpeg session was created for any failing attempt, while every successful one produced them normally. That absence is what makes this hard to spot from the server side.

H.264 sources were unaffected because their own level was already 4.0 or 4.1, so the server had no reason to go higher — which is why this looked like an HEVC problem.

## Related Issues
- Related to #1205
- Related to #1222

## Type of Change
- [x] Bug fix

## Changes Made
- Cap `VideoLevel` at 41 in the Chromecast codec profile
- Add regression tests for the level cap, the absence of High 10 from the profile list, and the h264-only transcoding profile
- Declare `@@locale` in `app_peo.arb`, the same one-line change as #1222, without which CI cannot reach this fix. Happy to drop that commit and rebase if #1222 lands first.

## Platform
- [x] Android
- [x] iOS
- [x] All / Shared code

## Testing

### Test Steps
1. Cast a 1080p HEVC title to a 3rd gen Chromecast on main: the app reports a receiver error within seconds and `config/log/` holds no FFmpeg.Transcode session for that item.
2. Apply the change and cast the same title: playback starts, and a transcode session appears and runs normally.
3. Repeated with a 2160p HEVC 10-bit TrueHD title, which also plays after the change and did not before.
4. `flutter test test/playback/chromecast_device_profile_test.dart` passes.

## Checklist


jmawet:
Please use the PR template

DrPleb:
> Please use the PR template

RadicalMuffinMan:
I merged main in and pushed a small correction to the branch. Two things:

The l10n commits are already on main. Commit f6186a92 there did both jobs at once: it fixed the Finnish `{tyyppi}` placeholder and deleted `lib/l10n/app_peo.arb` outright as an obsolete translation. That's also why #1222 was closed rather than merged.

The rationale needed correcting, though the fix itself is right. The comment said generations 1 through 3 stop at 4.1 and only Ultra and Google TV go past it. Google's published Cast media specs say otherwise:

https://developers.google.com/cast/docs/media#video_codecs

- Chromecast 1st and 2nd Gen: H.264 High Profile up to level 4.1
- Chromecast 3rd Gen: H.264 High Profile up to level 4.2
- Chromecast Ultra: H.264 High Profile up to level 4.2
- Chromecast with Google TV: H.264 High Profile up to level 5.1

So 3rd gen and Ultra both document 4.2 support. Why your 3rd gen device rejected `avc1.64002A` I genuinely can't explain, and I'm not disputing what you saw on real hardware.

The cap is still the right change, just for a different reason, so I reworded the comment rather than dropping it.

PR #1226, opened 2026-08-21
Feat/cast honour the user's max bitrate when casting

DrPleb:
# Pull Request

## Summary
Builds on the level 4.1 fix — please read that PR first, this branch is based on it.

The cast profile hard-coded 20 Mbps, while `aether_backend` and `appletv_backend` already read `UserPreferences.maxBitrate` for the local players. Casting was the one path where the user's own ceiling had no effect.

On a Chromecast 3rd gen that default is unreachable: playback starts and then stalls, and the receiver reloads the stream from the beginning every few seconds. Measured on the same file, network and server:

    5 Mbps   -> plays, one ffmpeg session
    10 Mbps  -> stalls, repeated sessions
    20 Mbps  -> stalls within three seconds

The server was not the limit — the transcode ran at 7-12x realtime with no errors — and neither was the network: a WAN with 20 Mbps upload and a 433 Mbps WiFi link behaved identically, and the same file plays fine on a SHIELD TV on that same network. The receiver simply cannot pull that much, and until now nothing in the app could lower it.

Unset or zero keeps the previous 20 Mbps default, so nothing changes for anyone not hitting this.

## Related Issues
- Related to #1205
- Related to #1222

## Type of Change
- [x] Bug fix
- [x] New feature

## Changes Made
- `chromecastDeviceProfile()` takes an optional `maxBitrateMbps` and falls back to the previous 20 Mbps default
- `GoogleCastProvider` reads `UserPreferences.maxBitrate`, the same preference the local backends use
- Add regression tests for the default and for the user-set ceiling
- Declare `@@locale` in `app_peo.arb`, the same one-line change as #1222, without which CI cannot reach this fix

## Platform
- [x] Android
- [x] iOS
- [x] All / Shared code

## Testing

### Test Steps
1. Set the max bitrate preference to 5 Mbps and cast a 2160p HEVC title to a 3rd gen Chromecast: the server is asked for 4616000 bps and playback runs uninterrupted.
2. Set it to 10 Mbps and cast the same title: the server is asked for 9616000 bps and playback stalls, with repeated transcode sessions in the server log.
3. Leave the preference unset: the profile still requests 20 Mbps, unchanged from before.
4. `flutter test test/playback/chromecast_device_profile_test.dart` passes.

## Checklist


jmawet:
Please use the PR template

DrPleb:
> Please use the PR template

RadicalMuffinMan:
Good catch on the underlying gap. One problem though, and it inverts the result. The description says "unset or
zero keeps the previous 20 Mbps default", but the preference is never unset: `UserPreferences.maxBitrate` has `defaultValue: '120'`. So on an untouched install the profile now asks for 120 Mbps where it previously asked for 20:

    maxBitrate default preference : 120
    parsed as Mbps                : 120
    cast MaxStreamingBitrate      : 120000000
    previous hardcoded value      : 20000000

That is a 6x increase handed to everyone by default, on the profile that deliberately targets the weakest Cast device, which is the same stalling failure this PR set out to fix. Users on `auto` are fine, since `int.tryParse('auto')` is null and falls back to 20.

The reasoning for capping rather than replacing is that the preference describes the user's network, not the receiver. Someone sets 120 because their LAN is fast, which says nothing about what their Chromecast can pull. It fits how the rest of this profile already works, capping width at 1920 and level at 4.1 because the sender cant ask which model it reached, with the preference as the escape hatch for a receiver weaker than that floor assumes.

PR #1227, opened 2026-08-21
fix(cast): list each Cast receiver once

DrPleb:
# Pull Request

## Summary
Every Cast device appears twice in the picker on Android 11 and later.

Android publishes each receiver through both provider services, so the same device arrives under two route ids that differ only in their prefix:

    com.google.android.gms/.cast.media.CastMediaRouteProviderService_Persistent:4d70b42e...
    com.google.android.gms/.cast.media.CastMediaRoute2ProviderService_Persistent:4d70b42e...

The part after the colon identifies the receiver. Selecting either entry always produced the same session, so nothing changes but the list.

## Related Issues
- Related to #1222

## Type of Change
- [x] Bug fix
- [x] UI/UX update

## Changes Made
- Key Cast routes on the receiver id rather than the full route id, so each device is listed once
- Apply it to both discovery paths: the snapshot and the event stream that feeds the picker
- Declare `@@locale` in `app_peo.arb`, the same one-line change as #1222, without which CI cannot reach this fix

## Platform
- [x] Android

## Testing

### Test Steps
1. On Android 12 with two Cast receivers on the network, open the cast picker on main: each device is listed twice.
2. Apply the change and reopen the picker: each device appears once.
3. Cast to the remaining entry and confirm the session starts as before.

## Checklist


jmawet:
Please use the PR template

DrPleb:
> Please use the PR template

PR #1257, opened 2026-08-23
fix(downloads): honor the server download permission, dispose the service on server switch, and respect the concurrency setting

RadicalMuffinMan:
enforced the download permission at login and in the track menu, disposed the download service on server switch, and honored the concurrent downloads setting


RadicalMuffinMan:
Closing, opened by mistake. This work belongs on #1253.

PR #1301, opened 2026-08-26
feat: optimize visual and settings performance on TV

sabino:
# Pull Request

## Summary

Add device-wide visual controls and a reduced-motion settings path for
low-power TV clients: a master backdrop switch, a TV-aware backdrop render
profile, a master switch that lets the operating-system screensaver take over,
and immediate settings-panel navigation on TV.

## Related Issues

None.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [x] Performance improvement
- [x] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

- Add unscoped device-wide backdrop and in-app screensaver master switches
  while preserving per-profile settings.
- Route home, classic/modern detail, Seerr, media-bar variants, prefetch, and
  background synchronization through the effective backdrop setting.
- Add Automatic, Quality, and Performance rendering modes; Automatic uses a
  720 px decode budget with no crossfade or Ken Burns motion on TV, while
  retaining the existing quality path elsewhere.
- Release the screensaver wake lock when the device master switch is disabled
  so the Android system dream can run.
- Add a shared settings motion profile. Android TV and accessibility
  reduced-motion clients open the settings panel and nested pages immediately,
  use hard-edge panel clipping, skip animated focus scrolling, and avoid the
  blurred focus glow. Non-TV clients retain the existing transitions and glow.
- Add English and Portuguese settings copy, generated localizations, settings
  search entries, and focused regression tests.

## Platform

- [x] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [x] All / Shared code

## Testing

### Test Steps

1. Ran focused preference, backdrop-render-profile, and settings-motion tests,
   app analysis, and debug/release Android TV builds in CI.
2. Installed the signed beta side by side on a 32-bit ARM Android 9 TV with
   2 GiB RAM; verified home, media bar, modern detail, settings, and nested
   settings pages with device backdrops disabled.
3. Verified the app held no wake lock or keep-screen-on window with the device
   screensaver switch disabled, then confirmed the configured Android photo
   dream took over during idle.
4. Captured two warmed single-DPAD-right home-row samples. The current stable
   control peaked at 127% CPU; the changed build repeated at 17.1% and 3.1%.
5. Measured the sidebar Settings panel separately. A warm pre-change open used
   467 app CPU ticks over 6.930 s (67.4% overall, 236.9% peak) and grew PSS by
   7,330 KiB. The normalized changed-build repeat used 60 ticks over 8.452 s
   (7.1% overall, 166.2% peak) and grew PSS by 1,615 KiB: about 87% fewer ticks,
   89% lower overall CPU, 30% lower sampled peak, and 78% less PSS growth.
6. Verified that a single focus move remained within measurement noise (it was
   already cheap), and that the nested Personalization page opens and navigates
   back correctly.

CI build and validation runs:

- https://github.com/sabino/Moonfin-Core/actions/runs/32994157088
- https://github.com/sabino/Moonfin-Core/actions/runs/33027182766

The measurements are device-level directional evidence, not general
benchmarks.

Manual 1920x1080 captures were reviewed locally. They contain a private media
library and profiles, so they are intentionally not attached.

## Checklist

### DIFF
--- a/lib/data/services/background_service.dart
+++ b/lib/data/services/background_service.dart
@@ -5,6 +5,7 @@ import 'package:flutter/painting.dart';
 import 'package:get_it/get_it.dart';
 
 import '../../data/models/aggregated_item.dart';
+import '../../preference/user_preferences.dart';
 import 'connectivity_service.dart';
 import 'media_server_client_factory.dart';
 
@@ -44,7 +45,12 @@ class BackgroundService {
   Stream<String?> get backgroundStream => _backgroundController.stream;
   Stream<BlurContext> get blurContextStream => _blurContextController.stream;
 
+  bool get _backdropsAllowed =>
+      !GetIt.instance.isRegistered<UserPreferences>() ||
+      GetIt.instance<UserPreferences>().shouldShowBackdrops;
+
   void setBackground(AggregatedItem? item, {BlurContext context = BlurContext.details}) {
+    if (!_backdropsAllowed) return clearBackgrounds();
     if (item == null) return clearBackgrounds();
 
     _blurContext = context;
@@ -130,6 +136,7 @@ class BackgroundService {
   }
 
   void setBackgroundUrl(String url, {BlurContext context = BlurContext.browsing}) {
+    if (!_backdropsAllowed) return clearBackgrounds();
     _blurContext = context;
     _blurContextController.add(context);
     _loadBackgrounds([url]);
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -3269,6 +3269,34 @@
   "@showBackdropImages": {
     "description": "Description for background backdrops"
   },
+  "deviceBackdrops": "Backdrops on This Device",
+  "@deviceBackdrops": {
+    "description": "Device-wide master setting for backdrop artwork"
+  },
+  "deviceBackdropsSubtitle": "Overrides every profile. Turn off to stop loading and displaying background artwork on this device.",
+  "@deviceBackdropsSubtitle": {
+    "description": "Description for the device-wide backdrop override"
+  },
+  "backdropRendering": "Backdrop Rendering",
+  "@backdropRendering": {
+    "description": "Setting for the backdrop rendering budget"
+  },
+  "backdropRenderingSubtitle": "Automatic optimizes TVs. Quality keeps full artwork and crossfades; Performance uses smaller artwork and instant changes to reduce lag.",
+  "@backdropRenderingSubtitle": {
+    "description": "Description for the backdrop rendering budget"
+  },
+  "backdropRenderingAutomatic": "Automatic",
+  "@backdropRenderingAutomatic": {
+    "description": "Backdrop rendering option: choose for this device"
+  },
+  "backdropRenderingQuality": "Quality",
+  "@backdropRenderingQuality": {
+    "description": "Backdrop rendering option: full artwork and transitions"
+  },
+  "backdropRenderingPerformance": "Performance",
+  "@backdropRenderingPerformance": {
+    "description": "Backdrop rendering option: reduced work"
+  },
   "seriesThumbnails": "Display Series Thumbnails",
   "@seriesThumbnails": {
     "description": "Setting for series thumbnails"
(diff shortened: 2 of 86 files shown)


mattsigal:
Just looking over this description, it seems like there are a variety of things here that either a) are already in Moonfin, or b) are not really within the scope of the PR.

Moonfin already has toggles to disable backdrops (in General Style for the home page and details pages, and under Library for browsing Libraries), as well as screensaver options (disabling Moonfin's screensaver should allow the system one to kick in). I am curious how your backdrop toggle differs from those to the point where additional settings would be necessary? 

If the OS screensaver isn't kicking in, that would be a bug and that code should be submitted as its own PR.

"Add Automatic, Quality, and Performance rendering modes; Automatic uses a 720 px decode budget with no crossfade or Ken Burns motion on TV, while retaining the existing quality path elsewhere." -- this seems like the crux/most interesting part of the PR. Your last bullet also seems to throw in a variety pack of things that could also be separated into more targeted PRs or personal adjustments (e.g., I'm not sure what "Portuguese settings copy" is doing here?).

RadicalMuffinMan:
@mattsigal brings up some great points @sabino

sabino:
@mattsigal @RadicalMuffinMan Thank you — I agree that the PR has accumulated too many independent concerns, and I appreciate the push to narrow it.

The motivation for the extra backdrop/screensaver switches was a shared TV with four local Jellyfin profiles: the existing preferences are profile-scoped, so repeating onboarding and configuration for each profile is tedious. That said, I agree this PR should not introduce a second policy layer to solve that workflow. I’m happy to remove both device-wide master switches and retain Moonfin’s existing per-profile behavior. If disabling Moonfin’s existing screensaver preference does not allow the OS screensaver to run, I’ll treat that as a separate reproducible bug.

A general “apply these device settings to all local profiles” workflow may be a better solution to the shared-TV problem. Is that something Moonbase/profile sync is intended to cover? If not, I can raise it as a separate discussion rather than adding per-feature overrides here.

While tracing the existing `backdropEnabled` preference, I found that some details/media-background paths do not appear to honor it consistently. Would you prefer that consistency fix as its own small PR from current `main`, or is the minimum propagation fix appropriate to keep in #1301?

For #1301, I propose retaining only the policy-neutral TV rendering/performance work and its tests: bounding backdrop decode/compositing cost and avoiding crossfade/Ken Burns work on the performance path. I can move the settings-panel performance changes into a separate PR created fresh from current `main`. For the backdrop optimization itself, would you prefer the current explicit Automatic/Quality/Performance choice, or an automatic TV optimization with no new user-facing setting?

I’ll also remove the direct Portuguese localization from this PR and handle translation separately through the project’s Weblate workflow.

If that split matches your expectations, I’ll update #1301 accordingly; any follow-up PRs will start from a freshly pulled current `main`.

RadicalMuffinMan:
Which low power tv device are you using that requires these changes?

RadicalMuffinMan:
But yes, thank you, this is a good split and it matches what we'd want.

On the shared-TV problem, yes pref_show_backdrop is in _scopedPreferenceKeys, so it is stored per server and user. The Moonbase plugin would handle the "apply to all local profiles" concept today. 

On the backdropEnabled inconsistency: own PR please, and it's narrower than you may think. The gap is media_bar.dart, which honors neither backdropEnabled nor hideBackdropsInLibraries. Everything else already checks one or the other. That should be a small, self-contained fix, and I suspect it removes most of what pushed you toward device-wide switches in the first place.

On the rendering setting: I don't think we need one. Blurred backdrops are already decoded small, so capping them at 720 on TV costs nothing you can see. Just do it. The media bar hero is the exception because it isn't blurred and fills the screen, so 1280 down to 720 would visibly soften the biggest image on the home page. Cap the blurred ones, leave the hero at 1280, and there's no new setting to add. If it turns out one is still needed we can add it later.

One thing on the benchmark. 127% CPU dropping to 17.1% and 3.1% is too big a swing for a decode-width change on its own, and it usually means the control run was doing first-time decodes and downloads while the changed run was warm. I ran into exactly that testing artwork changes on a Fire TV Stick 4K 1st Generation a couple of hours ago, where the build that was genuinely lighter looked slower, purely because the server had never encoded the sizes it was being asked for. Worth re-running both sides fully warm before we lean on the numbers.

Also a heads up that main moved under you. 238b70af and 6ad150ce changed how artwork is decoded and requested, including cacheWidthFor in bounded_network_image.dart, which is what your maxDecodeWidth ends up feeding. Worth rebasing and taking the measurement again from current main.

PR #1402, opened 2026-09-04
build(ios): override code signing in `build-ios.sh`

jmartindf:
# Pull Request

## Summary
The `build-ios.sh` script isn't currently used anywhere and was giving me errors around signing and entitlements when I tried to use it. I've updated the script to allow signing with a developer's own Apple Development ID, overriding entitlements the development doesn't have, and overriding the display name to distinguish between an App Store Moonfin install and a development Moonfin install on the same device.

## Related Issues
N/A

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [x] Build/CI change
- [ ] Other (describe):

## Changes Made
Using `build-ios-private.env`:

- Override the display name with `OVERRIDE_APP_NAME=Sidefin`
- Override the Signing Development Team with `OVERRIDE_DEVELOPMENT_TEAM=ADevTeamID`
    - Also overrides the Bundle Identifier to `ADevTeamID.org.moonfin.app`, to match a developer's wildcard iOS development profile.

The default app name and bundle identifier are now configured in `ios/Flutter/Shared.xcconfig`, instead of directly into the Xcode project's GUI interface.

## Platform
- [ ] Android
- [x] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] All / Shared code

## Testing
I did all of the below steps, except for \#2, without issues.

### Test Steps
1. Run `./build-ios.sh` and verify that an unsigned IPA is built.
2. Edit `build-ios-private.env`, setting `IOS_CODESIGN="1"`, and verify that a signed App Store IPA is built (only if you're on the 'CCBYLPUSVH' development team).
3. Edit `build-ios-private.env`, setting `IOS_CODESIGN="1"`, `IOS_EXPORT_METHOD="development"`, `OVERRIDE_APP_NAME=Sidefin`, and `OVERRIDE_DEVELOPMENT_TEAM=F6J3WTU2J9`, and verify that a signed IPA is built (only if you're on the 'F6J3WTU2J9' development team) and ready to be loaded onto one of your iOS devices.

N/A

## Checklist


RadicalMuffinMan:
Hey bud, thanks for the PR and for testing with a real device. build-ios.sh isn't actually "unused", it's used for every GitHub and Apple App Store releases and I also use it to build local files for testing. The blocker is the `IOS_CODESIGN` branch. The script runs `set -euo pipefail` and line 114 reads `$IOS_CODESIGN` without a default, so with the variable unset the build dies with unbound variable. Since build-ios.private.env is gitignored, that's what anyone gets on a fresh clone, and it fails only after the cores build and pod install have run.

Past that, the branch makes the two builds exclusive where today they both come out of one run. I'd have to run the whole script twice to get what one run gives me now, paying for the cores and pods each time. Could you default it `(IOS_CODESIGN="${IOS_CODESIGN:-0}")` and keep both builds unconditional with the flag skipping just the signing step? That keeps one run producing both and still gives someone without a signing identity a clean unsigned-only run.

Two other small things: the bundle id built from a Team ID is unusual enough to deserve a note in the example env, and the overrides only apply on the `--export-method` branch so anyone using `IOS_EXPORT_OPTIONS_PLIST` silently gets none of them.

jmartindf:
Thanks for taking a look and for the feedback! Sorry about the undefined variable path/crash.

Changes:

1. Out of the box, the script creates **_only_** an unsigned build in "Moonfin_iOS_v2.5.1_unsigned.ipa"
2. Setting `IOS_CODESIGN="1"` in build-ios.private.env causes the script to create an unsigned build and then a signed build, with the default signing identity, name, bundle identifier, and `app-store` export method.
3. Setting `IOS_EXPORT_METHOD="development"` changes only the export method from change number 2.
4. Setting `OVERRIDE_APP_NAME=Sidefin` changes only the display name for both created IPAs.
5. Setting `OVERRIDE_DEVELOPMENT_TEAM=ADevTeamID` changes both the signing identity and the bundle identifier.
6. The overrides now apply for both the `IOS_EXPORT_METHOD` and the `IOS_EXPORT_OPTIONS_PLIST` branches. (Although I'm not sure how well `IOS_EXPORT_OPTIONS_PLIST` works, given my own testing yesterday + https://github.com/flutter/flutter/issues/113977).

(I'm not sure if it makes sense to override the app name if you're not also overriding the bundle identifier, but I'm leaving it as-is for now.)

Can you take another look and let me know if this makes the updated script usable for you? If so, I'll do one more update afterwards, to document everything in the example env.

RadicalMuffinMan:
Thanks for the quick turnaround and for catching the `IOS_EXPORT_OPTIONS_PLIST` branch properly rather than just the one I pointed at.

Before you spend time on the example env docs though, hold off for a second. I never documented how to build for iOS anywhere, so you reasonably picked up the script. But `build-ios.sh` is what cuts every release, and for testing you don't need it at all: open `ios/Runner.xcworkspace` in XCode or run `flutter run`. The simulator does no code signing, so nothing in this PR is required there.

Where you ran into an issue is testing on a physical device,and it isn't the script. The app ships two entitlements Apple only grants on request:

```
com.apple.developer.networking.multicast
com.apple.developer.carplay-audio
```

A normal account cant sign with either so a device build fails even once the team and bundle identifier are right. Nothing in the repo said so which is the actual bug here. I've just added an iOS section to the Building from Source wiki page covering it. So the problem is real and worth fixing, just not through the release script. What I'd rather have is a local override file that Xcode and `flutter run` both pick up on their own, with no env vars and no script involvement.

Two pieces of what you've built are exactly that mechanism and I want to keep them:

- `Shared.xcconfig` with the bundle identifier and display name templated
- `RunnerSideload.entitlements`

What's missing is the other two settings, and they're the ones that matter most, because they're what Xcode writes into `project.pbxproj` when you change them in the UI. That tracked file getting dirty is the thing a contributor actually trips over:

```
// Shared.xcconfig
MOONFIN_BUNDLE_IDENTIFIER = org.moonfin.app
MOONFIN_BUNDLE_DISPLAY_NAME = Moonfin
MOONFIN_DEVELOPMENT_TEAM = CCBYLPUSVH

#include? "Local.xcconfig"
```

with `DEVELOPMENT_TEAM` and `CODE_SIGN_ENTITLEMENTS` templated in the pbxproj the same way you did the bundle identifier. The entitlements default differs per configuration, so that one belongs at the top of `Debug.xcconfig` and `Release.xcconfig` rather than in `Shared.xcconfig`.

The `#include?` goes last so a gitignored `Local.xcconfig` wins, and a contributor then writes only:

```
MOONFIN_DEVELOPMENT_TEAM = THEIRTEAMID
MOONFIN_BUNDLE_IDENTIFIER = com.theirname.moonfin
MOONFIN_ENTITLEMENTS = Runner/RunnerSideload.entitlements
```

Nothing tracked changes, `build-ios.sh` reverts to what it is on main, and the `IOS_CODESIGN` question goes away entirely.

Also you're right, different display name on its own doesn't buy anything, because two apps sharing a bundle identifier can't both be installed anyway. It only makes sense alongside an overridden identifier, so tying it to the same local file is the natural place for it.

One small thing if you do keep any of the script work: the default came out as `IOS_CODESIGN="${IOS_CODESIGN:-\"0\"}"`, which makes it the string `"0"` with the quotes in it. That's the snippet I gave you minus the escaping, my fault for pasting it inside parens. It behaves fine against the `!= "1"` test so it wouldn't show up in testing.

And one more thing you'd want to know regardless: Profile has no xcconfig of its own and inherits `Release.xcconfig`, so anything you template needs to resolve there too. Easy to miss, and it fails as an empty bundle identifier rather than an error.

Happy to take this reshaped if you're up for it. If you'd rather hand it over at this point that's completely fair, you've already done the hard part by finding the entitlements wall. Thank you!

jmartindf:
I'll reshape this.

jmartindf:
Ready for review again.

As suggested, the bulk of the changes are now in templated values and can be overridden in a gitignored Local.xcconfig. I provided a Local.xcconfig.example that shows what to override, why, and how.

> Nothing tracked changes, `build-ios.sh` reverts to what it is on main, and the `IOS_CODESIGN` question goes away entirely.

The current build-ios.private.env.example mentions using `IOS_CODESIGN="0"` to skip making a signed IPA, but the current script doesn't actually support that. I've left in the update to build-os.sh so that it respects `IOS_CODESIGN="0"`, building the unsigned IPA and skipping the signed IPA entirely.

> One small thing if you do keep any of the script work: the default came out as `IOS_CODESIGN="${IOS_CODESIGN:-\"0\"}"`, which makes it the string `"0"` with the quotes in it. That's the snippet I gave you minus the escaping, my fault for pasting it inside parens. It behaves fine against the `!= "1"` test so it wouldn't show up in testing.

Fixed.

> And one more thing you'd want to know regardless: Profile has no xcconfig of its own and inherits `Release.xcconfig`, so anything you template needs to resolve there too. Easy to miss, and it fails as an empty bundle identifier rather than an error.

I think this works?

PR #1406, opened 2026-09-04
Fix deep-link warm re-nav auto-play and add optional userId pin

spsDrop:
Pull Request

## Summary
I was setting up some home automation to allow for directly playing titles over the Home Assistant android remote protocol and found deeplinking had some bugs and the user select screen could not be bypassed via deeplink even though I knew which profile would be watching.

Two fixes for `moonfin://play` deep links on Android TV.

Warm re-navigation (app running, fire a link to another item) landed on the detail page and never played, and after that every play press in the process silently did nothing. Root cause: `launchPlayerWhilePreparing` holds a process-wide video-launch slot across `await routeFuture`; a deep link navigates via `go()`, which removes the player route declaratively, and go_router only completes push futures through the pop path — so the slot was never released. The fix releases the slot when the holder's `/player/video` route is no longer on the stack (a live holder always has it there).

Cold start deep links still hit the profile switcher. This adds an optional `userId` deep-link param: o> so that items can be played directly on profiles. This skip only occurs if the useId is provided in the deeplink and there is no pin requirement.

Also adds unit tests for `routeForDeepLink`, which had none.

## Related Issues
None filed — happy to add one first if maintainers prefer.

## Type of Change
- [x] Bug fix
- [x] New feature (optional `userId` deep-link param)

## Changes Made
- `lib/ui/navigation/playback_launcher.dart`: release a stale video-launch slot when the holder's player route was removed by a location change without completing its push future. Such a slot held the process-wide launch gate for the rest of the app's life, swallowing every later play press.
- `lib/data/services/deep_link_service.dart`: forward an optional `userId` query param into the item route (alongside `serverId`).
- `lib/ui/navigation/deep_link_navigator.dart`: on cold start with `userId`, resolve the server (explicit `serverId` or unique match across stored servers, never guessed), run `loadStoredServers()` + `switchCurrentSession()`, then navigate. Skipped when the user has a PIN or `shouldAlwaysAuthenticate` is set; falls back to the picker path on any failure.
- `test/data/services/deep_link_service_test.dart`: unit coverage for `routeForDeepLink`.

## Platform
- [x] All / Shared code (behavior observed on Android TV; changes are in shared `lib/`)

## Testing
- [x] Tested on physical device (Android TV box)
- `flutter analyze` clean on changed files; unit tests pass.

### Test Steps
1. Fresh install, cold start: fire `moonfin://play?id=<episode>&userId=<stored user>` — profile picker is skipped and the episode auto-plays as that user.
2. While playing, fire `moonfin://play?id=<different episode>` — the new episode auto-plays (before: detail page only, then every play press in the process dead).
3. Repeat step 2 several times in one session — all play (before: locked after the first).
4. Cold start with a `userId` that has a PIN enabled (or always-authenticate on) — the picker still appears; the pin is not bypassed.

None — behavioral fix, no UI changes.

## Checklist
- [x] Code builds successfully (`flutter build apk --debug --flavor androidTv`)
- [x] No new warnings introduced (`flutter analyze` clean on changed files)

### DIFF
--- a/lib/data/services/deep_link_service.dart
+++ b/lib/data/services/deep_link_service.dart
@@ -39,15 +39,22 @@ class DeepLinkService {
   /// the link isn't recognized.
   ///
   /// Recognized forms:
-  ///  - `moonfin://item?id=<itemId>[&serverId=<serverId>]`
-  ///  - `moonfin://play?id=<itemId>[&serverId=<serverId>]` (starts playback)
+  ///  - `moonfin://item?id=<itemId>[&serverId=<serverId>][&userId=<userId>]`
+  ///  - `moonfin://play?id=<itemId>[&serverId=<serverId>][&userId=<userId>]`
+  ///    (starts playback)
+  ///
+  /// `userId` optionally pins a stored user for cold starts, skipping the
+  /// profile picker. It is only honored for users already signed in on this
+  /// device, and never bypasses a PIN or an always-authenticate setting.
   static String? routeForDeepLink(Uri uri) {
     if (uri.scheme != 'moonfin') return null;
     final id = uri.queryParameters['id'];
     if (id == null || id.isEmpty) return null;
     final serverId = uri.queryParameters['serverId'];
+    final userId = uri.queryParameters['userId'];
     final params = <String, String>{
       if (serverId != null && serverId.isNotEmpty) 'serverId': serverId,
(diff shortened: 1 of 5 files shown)


RadicalMuffinMan:
pull latest from main when you get a chance, thank you

RadicalMuffinMan:
The stale slot check asked the app-wide router instead of the one the launch had pushed onto.

PR #1429, opened 2026-09-06
[style]: editorconfig

jmartindf:
# Pull Request

## Summary
Bundle a `.editorconfig` file, so that well-behaved code editors will automatically use Moonfin's preferred styles for tabs-vs-spaces and indent width.

(Because my own defaults are different and I have to keep remembering to manually change the settings and why put myself through that?)

## Related Issues
N/A

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [x] Other (describe): developer experience

## Changes Made
A bundled `.editorconfig` file specifying:

- Unix-style line endings
- no final newlines at the end of files
- trim trailing whitespace from each line
- UTF-8 character set
- indent with spaces, not tabs
- indent width is 2 characters

## Platform
- [ ] Android
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [x] All / Shared code

## Testing

### Test Steps
1. Open files in Neovim. Verify that indenting / unindenting, etc all match Moonfin's existing files, rather than my own personal formatting preferences

N/A

## Checklist
- [ ] Code builds successfully
- [ ] No unnecessary commented-out code
- [ ] No new warnings introduced

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,24 @@
+	# Tell text editors what Moonfin's preferred coding styles are
+# https://EditorConfig.org
+
+# Unix-style newlines, with a newline ending every file, utf-8 charset
+# default to spaces not tabs, with 2 spaces for each indent
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+indent_style = space
+indent_size = 2
+
+# Native platform sources indent with four
+[*.{kt,kts,java,swift,gradle,xml,cpp,cc,h}]
+indent_size = 4
+
+# Xcode owns the formatting of these
+[*.{pbxproj,plist,entitlements,xcscheme,xcconfig}]
+indent_style = tab
+
+# Vendored upstream code keeps its own line endings
+[packages/flutter_inappwebview_windows/**]
+end_of_line = crlf


RadicalMuffinMan:
Thanks for bundling an editorconfig, its a good idea and there's no existing one to conflict with.

Two of the settings would fight the tree though and one of them you've already fixed by hand elsewhere.

`insert_final_newline = false` is the main one. `dart format` always ends a file with a newline, so an editor stripping it just means the formatter puts it straight back. 388 of the 398 text files I checked already end with one. Your own #1402 adds the missing final newline to `build-ios.sh`, which is the same thing from the other side. This should be `true`.

The blanket `indent_size = 2` is right for Dart and YAML but not for the rest of the tree. Kotlin, Swift, Gradle and XML are all on 4 here. The Xcode files are tab indented, `project.pbxproj` alone has 835 tab indented lines and no space indented ones, so pointing editors at spaces there means Xcode rewrites them back the next time anyone changes a setting in the UI. 

Also worth adding `root = true` at the top otherwise editorconfig keeps walking up past the repo and picks up whatever a parent directory happens to define.

Something like this matches what's actually in the tree:

```
root = true

[*]
end_of_line = lf
insert_final_newline = true
trim_trailing_whitespace = true
charset = utf-8
indent_style = space
indent_size = 2

# Native platform sources indent with four
[*.{kt,kts,java,swift,gradle,xml,cpp,cc,h}]
indent_size = 4

# Xcode owns the formatting of these
[*.{pbxproj,plist,entitlements,xcscheme,xcconfig}]
indent_style = tab

# Vendored upstream code keeps its own line endings
[packages/flutter_inappwebview_windows/**]
end_of_line = crlf
```

`trim_trailing_whitespace` and `charset` are both good as they stand. I checked and none of the markdown files rely on trailing double spaces for hard breaks so nothing gets hurt there.

Thank you!

jmartindf:
Made the suggested changes, but leaving out the `root = true` line. 

> Properties from matching EditorConfig sections are applied in the order they were read, so properties in closer files take precedence.

By allowing editors to find editorconfig files farther up, users can get their own personalized settings for `tab_width` in Xcode files that use tab indenting.

PR #1486, opened 2026-09-10
AV1 Dovi 10.1 throwaway Dovi

MasterOfKay:
# Pull Request

## Summary
AetherEngine hase problems with Dovi 10.1 play. Transcoding also fails as Aether engine marks the dovw files as dav1. Dav1 can't be translated by ffmpeg wich results to the Error Transcoding failed. Or in rare cases it wires through as a SDR transcode.

So if a file is AV1 10.1 Dovi but hase a hdr layer we throw dovi away and play hdr as direct play withc works better.

## Related Issues

- Closes #
- Fixes #
- Related to #

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

- Ather bakend
- 
- 

## Platform
- [ ] Android
- [ ] Android TV
- [X] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] All / Shared code

## Testing

### Test Steps
1. Loaded new ipa
2. Played AV1 Doviwwithhdr10plus file and saw it direct playes now
3.

Bevor the changes:
WhatsApp Image 2026-09-10 at 23 19 15 (1)
WhatsApp Image 2026-09-10 at 23 19 15

After the changes:
WhatsApp Image 2026-09-10 at 23 19 16

## Checklist


RadicalMuffinMan:
Had a look through this. The bug is real, I checked: AV1 Profile 10.1 routing hasn't changed at all between the engine 2.5.1 shipped and 6.75.0, which is what main is on now. Everything that did change in `CodecRoutePolicy` in that range is HEVC Profile 8 work. So this isn't going to fix itself in the next bump.

`preferredDecodePath` is also the right lever, it for updated in 6.75.0 for exactly this kind of thing.

A few things need sorting before this can go in.

 16 of the 19 files are `tizen/obj/`, about 1015 of the 1191 added lines, including an 886 line `project.assets.json`. Drop those. `tizen/obj` isn't in `.gitignore` either, which is how they got here, so adding it would be a nice bonus. Tizen was removed from the repo altogether.

The device profile change breaks other platforms. This bit:

```dart
       if (!supportsAv1Hdr10) {
         unsupportedRangeTypesAv1.add('DOVI_INVALID');
-      }
-      if (!supportsAv1Hdr10Plus) {
         unsupportedRangeTypesAv1.add('DOVI_WITH_HDR10_PLUS');
       }
```

moves `DOVI_WITH_HDR10_PLUS` onto the HDR10 gate, so any client with AV1 HDR10 but no HDR10+ now gets offered it for direct play. That includes Android via media3, which has no `preferSoftwareDecode` to fall back on. The device profile is shared by every backend, the fix should be for Aether only.

Your test asserts the new behavior so I assume this was deliberate, but it needs to be conditional on something the Aether backends set rather than applied everywhere.

Minor thing in the same area: the comment above `DOVI_INVALID` now sits above two adds and reads like it covers both.

tvOS isn't wired. `AppleTvVideoChannel` builds its `SourceConfiguration` without the new param, so only the iOS/macOS channel passes it. Compiles fine because the field defaults to false. Doesn't matter today since Apple TV has no hardware AV1 and already goes software, but it'll bite eventually.

The fallback promises more than the probe. The `main.dart` block is the fallback for when the native probe fails and it claims `maxResolutionAv1: 3840x2160` while both real probes say 1920x1080. It should match. The rest of that is already redundant anyway, both probes report every `supportsAv1*` key.

Your test header says FFmpeg's mp4 muxer has no `dav1` entry so `write_header` fails with EINVAL. I couldn't confirm that and there's a comment in the engine's muxer pointing the other way:

> `dvh1` is not in the mp4 table at all (mov.c handles it), so a Dolby Vision sample entry would be silently demoted to `hvc1`

`dvh1` isn't in the table either and HEVC DV muxes fine, because movenc takes the fourCC from `codec_tag` and the config box from `codec_id`. Same path should work for `dav1`.

Not saying you didn't hit something real, but the cause might be elsewhere and that changes what the fix should be. Can you attach the engine log from a failing play? The actual error line would sort it out.

Worth knowing too: `VideoRoutingPolicy` already returns `!av1Available` for AV1, and `VTCapabilityProbe.av1Available` is false on most Apple hardware through AVPlayer HLS. On those devices AV1 never touches the fMP4 muxer, so `preferSoftwareDecode` does nothing. It only changes anything on newer AV1 capable hardware on iOS 26.2+, which is presumably what you're on.

Longer term this belongs in the engine. It picks the `dav1` tag and knows whether it can mux it, so having every client detect Profile 10.1 in Dart means everyone reimplements the same rule and gets to be wrong separately. I'll raise it upstream once we know the actual failure. 

Also the description says the change makes it "play hdr as direct play", but forcing software decode is CPU decoding, not direct play. Update it so nobody reads this later during an Apple TV performance investigation and gets the wrong idea. The header comment on your test file is a much better description than the PR body,I'd just use that.

MasterOfKay:
Yes i called it to quickly a day. Here the Most important fromm all the logs wich i will attach in fully also:

From the Moonfin iOS log the failing play lines 776–805:

20:59:15.668650 [mp4 @ 0x11ad5cf00] [error] Could not find tag for codec av1 in stream #0,
                                            codec not currently supported in container
20:59:15.711523 [NativeAVPlayerHost] #5 item.status=failed err=AVFoundationErrorDomain/-11848 'Cannot Open'
20:59:15.714757 [AetherEngine] AVPlayer rejected the master (code=-11848); falling back to media playlist
20:59:15.820129 [mp4 @ 0x11ad5f700] [error] Could not find tag for codec av1 in stream #0, …
20:59:15.820134 [HLSSegmentProducer] muxer alloc for seg-0 failed: MP4SegmentMuxer: avformat_write_header failed (-22)
20:59:15.821452 [HLSVideoEngine] #99 VOD muxerFailed revive cap reached (3 failures, cap 2); giving up
20:59:15.821480 ERROR Native player error kind=unsupported_container recoverable=true

520 : av01.0.12M.….16.09.0,ec-3 -> direct played
733 : dav1.10.10,ec-3 -> muxer EINVAL, then -11848 

From the ffmpeg server logs the fallback that failure triggers:

-codec:v:0 libx265 -preset ultrafast -crf 28 tonemapx=tonemap=bt2390:t=bt709:m=bt709:p=bt709

Software x265 with an HDR→SDR tonemap. In the 09-10 run that was 11–18 fps against 23.976 needed, and Jellyfin restarted the encoder eleven times. Why i don't really know too.

I tried something now wich should actuall fix the case without breaking the other client parts. But i need to test this first now

Also i am sorry kind sir for calling a software decode direct play, as it kinda is for me. The Server doesn't need to touch the file. So this was direct enough for me.

[FFmpeg.Transcode-2026-09-12_19-28-24_7ead9c4718eca48125d28d4469f3a779_c3c6f0f7.log](https://github.com/user-attachments/files/32150929/FFmpeg.Transcode-2026-09-12_19-28-24_7ead9c4718eca48125d28d4469f3a779_c3c6f0f7.log)
[FFmpeg.Transcode-2026-09-12_19-28-32_7ead9c4718eca48125d28d4469f3a779_ad121fe5.log](https://github.com/user-attachments/files/32150930/FFmpeg.Transcode-2026-09-12_19-28-32_7ead9c4718eca48125d28d4469f3a779_ad121fe5.log)
[FFmpeg.Transcode-2026-09-12_19-28-37_7ead9c4718eca48125d28d4469f3a779_02a5207c.log](https://github.com/user-attachments/files/32150931/FFmpeg.Transcode-2026-09-12_19-28-37_7ead9c4718eca48125d28d4469f3a779_02a5207c.log)
[FFmpeg.Transcode-2026-09-12_19-29-09_7ead9c4718eca48125d28d4469f3a779_0a81dc5e.log](https://github.com/user-attachments/files/32150932/FFmpeg.Transcode-2026-09-12_19-29-09_7ead9c4718eca48125d28d4469f3a779_0a81dc5e.log)
[FFmpeg.Transcode-2026-09-12_20-09-17_902d3f422e6ffa253a6935273fb72361_5dbb96a2.log](https://github.com/user-attachments/files/32150933/FFmpeg.Transcode-2026-09-12_20-09-17_902d3f422e6ffa253a6935273fb72361_5dbb96a2.log)
[FFmpeg.Transcode-2026-09-12_20-09-48_902d3f422e6ffa253a6935273fb72361_bf47d78e.log](https://github.com/user-attachments/files/32150934/FFmpeg.Transcode-2026-09-12_20-09-48_902d3f422e6ffa253a6935273fb72361_bf47d78e.log)
[upload_Moonfin for iOS_2.5.1_20260910190046_0bb1157a1f6f4f9f916926323f428bae.log](https://github.com/user-attachments/files/32150935/upload_Moonfin.for.iOS_2.5.1_20260910190046_0bb1157a1f6f4f9f916926323f428bae.log)

MasterOfKay:
Ah shit right it pulls it directly here. Man i forgot that

MasterOfKay:
image

Well it still direct plays of the software decode. With now just applying to aether engin. Looked closer. Jellyfin native player muxes it to mp4 and plays the hdr file only. So this direct streaming without hitting the server seems for me a better traid off. But thats taste i guess. Apple can't play dv in a av1 file as it seemes. Well they can but over a super hacky way. And with ather engin in place already this seems like a solution for now till the engine or apple fixes av1 dv playing. Luckily hdr10 works. Hdr10plus will never be played as plus as apple can't do anything with the extra informations (on iPhone, newer appleTV's should play that). So hdr10 is is throughout.

Issue #1523, opened 2026-09-12
[Feature] Rename Auto Crop zoom mode — it is cover-zoom, not a crop


selmant:
### What problem does this solve?

Player zoom mode **Auto Crop** is cover-zoom: it scales the coded frame to fill the window and clips overflow (`BoxFit.cover` / `resizeAspectFill`). It does **not** detect or remove encoded letterbox bars. The name already oversells what it does.

### Proposed solution

Copy-only rename of the zoom-mode **label and description** so it is clearly fill/cover zoom, not bar detection.

Do not change zoom behavior.

Surfaces that say Auto Crop today:

- `lib/l10n/app_en.arb` — `autoCrop`, `settingsPlayerZoomDescription`
- Settings → Video playback → Player zoom mode
- Player OSD zoom cycle / toast (`l10n.autoCrop`)
- tvOS `ZoomMode` raw value / `displayName` (`"Auto Crop"`)
- iOS / macOS zoom wire (`autoCrop`, `Auto Crop`)

Leave the `ZoomMode.autoCrop` enum name unless a follow-up wants a code rename. English source first; Weblate already translates `autoCrop`.

Agree the new English label before shipping it.

### Platform

All / Not platform-specific

### Alternatives considered

- Keep Auto Crop. The current name is already wrong.
- Rename the enum (`autoCrop` → `fill` / `cover`). More churn than this issue needs.

### Additional context

Related closed: #858 (Auto Crop preference not applying). That was behavior, not naming.

PR #1524, opened 2026-09-12
Add encoded letterbox crop for desktop and Android

selmant:
## Summary
Adds a **Crop black bars** playback toggle that detects encoded letterbox/pillarbox and crops it, then fills the screen. This is a separate setting from zoom **Auto Crop** on purpose.

[#1523](https://github.com/Moonfin-Client/Moonfin-Core/issues/1523) is why: **Auto Crop is cover-zoom** (`BoxFit.cover` / aspect-fill). It scales the whole coded frame to fill the window and clips overflow. It does **not** find or remove black bars baked into a 16:9 file. Putting real bar-detection into that zoom control would make the existing confusion worse, so this ships as its own toggle next to zoom until that naming is sorted. If #1523 lands (rename Auto Crop so zoom is clearly fill/cover, not a crop), this can move into zoom settings.

## Related Issues
- Related to #1523

## Type of Change
- [x] New feature

## Changes Made
- Desktop libmpv: `cropdetect` then `vf crop`
- Android Media3 (phone and TV): PixelCopy (TextureView `getBitmap` fallback), then layout zoom into the crop rect (SurfaceView + tunneling cannot use Media3 Crop effects)
- Shared `LetterboxCropper` in `playback_core`; Aether / HTML / tvOS are explicit stubs
- Setting is hidden where no cropper runs

## Platform
- [x] Android
- [x] Android TV
- [x] Windows
- [x] Linux
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS

No devices on hand for iOS / macOS / tvOS / web. The cropper interface is there so those platforms can follow this path.

## Testing

### Test Steps
1. Settings → Playback & SyncPlay → Video Playback Preferences → enable **Crop black bars** (off by default)
2. Play a title with encoded letterbox (not just window pillarboxing)
3. After a few seconds the bars should crop and the picture should fill
4. Confirm zoom **Auto Crop** is still cover-zoom and is not this detect

Tested on Linux (libmpv) and Android TV emulator (Media3). Phone uses the same Media3 path; iOS/macOS/tvOS/web untested.

## Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/ios/Runner/Playback/AetherVideoPlatformView.swift
+++ b/ios/Runner/Playback/AetherVideoPlatformView.swift
@@ -91,7 +91,7 @@ final class AetherVideoPlatformView: NSObject, FlutterPlatformView {
     private static func zoomMode(fromWire value: String) -> ZoomMode? {
         switch value {
         case "fit", "Fit": return .fit
-        case "autoCrop", "Auto Crop", "Fill": return .autoCrop
+        case "autoCrop", "Auto Crop": return .autoCrop
         case "stretch", "Stretch": return .stretch
         default: return nil
         }
--- a/ios/Runner/Playback/AetherVideoPlatformView.swift
+++ b/ios/Runner/Playback/AetherVideoPlatformView.swift
@@ -91,7 +91,7 @@ final class AetherVideoPlatformView: NSObject, FlutterPlatformView {
     private static func zoomMode(fromWire value: String) -> ZoomMode? {
         switch value {
         case "fit", "Fit": return .fit
-        case "autoCrop", "Auto Crop": return .autoCrop
+        case "autoCrop", "Auto Crop", "Fill": return .autoCrop
         case "stretch", "Stretch": return .stretch
         default: return nil
         }
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -4124,15 +4124,15 @@
   "@fit": {
     "description": "Zoom mode: fit"
   },
-  "autoCrop": "Fill",
+  "autoCrop": "Auto Crop",
   "@autoCrop": {
-    "description": "Zoom mode: fill the screen (cover), clipping overflow. Not black-bar detection."
+    "description": "Zoom mode: cover the screen, clipping overflow. Not black-bar detection."
   },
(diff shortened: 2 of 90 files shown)


mattsigal:
Hope you don't mind me popping in to ask a few questions @selmant -- I just don't have a file at the ready to test this with and was curious.
1. If I understand this right, using the new toggles means it'll apply the crop just to the bars then do the requested zoom mode? So having this on + zoom mode set to Fit would mean it'll always show the entire image, yeah? And only make a difference if the source file has the black bars included.
2. Does this have any impact on direct playing? If the toggle kicks in, would that force a transcode?
3. There are some movies (e.g. films with some scenes that use IMAX dimensions) that switch between letterboxed and full-screen images. How would the toggle handle those cases?

selmant:
1. Not exactly. This is basically separate from the zoom mode. The zoom modes are mainly about how the video frame is fitted/stretched to the screen, while this specifically detects black bars that are actually encoded into the video and removes them. This is especially useful on ultrawide monitors. I have a 21:9 monitor myself. A lot of 18:9 / 21:9 movies are still encoded inside a 16:9 frame with black bars, so just using the existing zoom modes doesn't really solve the problem properly. Originally I actually disabled the normal zoom mode while this was enabled because once the bars are removed I didn't really need it anymore, but @RadicalMuffinMan changed that so the two can now be used independently. Maybe he has a different plan for how they should interact.

2. No, this doesn't modify the file on the server or force a transcode. Everything happens locally in the client. On desktop it's basically `cropdetect` + `vf crop`. The only slightly special part is hardware decoding. `cropdetect` needs access to the decoded frames, so if the current hwdec mode doesn't allow that, Moonfin temporarily switches to a compatible decoding mode while detecting the bars and restores it afterwards. So Direct Play itself is still Direct Play.

3. Right now I implemented it as a one-time detection near the beginning of playback, so movies that switch between letterboxed and IMAX/full-frame scenes won't update the crop dynamically. Maybe we can make it re-scan every 20 seconds or something like that. I think it's worth testing to see if that works well without causing visible jumps or too much overhead.

PR #1512, opened 2026-09-12
Feat: live tv tuner status

KevOutIngs:
# Pull Request

## Summary
Live TV channels that fail on the tuner side (Dispatcharr behind Jellyfin) used to leave the viewer on a black screen or a spinner with nothing to press, because the server never says whether the tuner is retrying or has given up. This PR adds a tracker that turns the bringup phases, buffering, the playback clock, the session end and (on Android) the rendered picture into a status: connecting, still trying, playing, buffering, reconnecting, unavailable or lost. Both live players show a spinner with a line for the waiting states and a Retry / Back card for the two failures. On tvOS the waiting states appear as a label over the native player and a failure dismisses the player so the card is what the viewer sees.

## Related Issues

- Closes #
- Fixes #
- Related to #

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Performance improvement
- [x] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

- New `LiveTvStreamStatusTracker` (pure, fixed-clock testable) and `LiveTvStreamStatusMonitor` in `lib/playback/live_tv_stream_status.dart`, with 33 unit tests taken from real tvOS, macOS and Android TV logs.
- `LiveTvStreamStatusOverlay`: spinner with a message for the waiting states, Retry / Back card for `unavailable` and `lost`, wired into `live_tv_player_screen.dart` and `appletv_livetv_player_host_screen.dart`.
- `PlaybackManager`: a live resolve that throws, or answers with the placeholder Jellyfin returns after a failed probe, becomes a failed bringup under `liveChannelUnavailableError` and the probe's live session is released; new `pictureShown` state and stream.
- `PlayerBackend.pictureShownStream` (optional). The Media3 view samples its surface for a black picture while a live channel plays, so a failover placeholder (black video with a sound track, which ExoPlayer plays normally) no longer counts as playing.
- Playing is declared as soon as a well-behaved clock steps after the stream opens; "still trying" is counted from the stream opening with an eight-second allowance that stops counting once the clock steps.
- tvOS native: tuner status shown as a label instead of an alert (the alert swallowed the Menu press); the player is dismissed from its presenter so a dismissal with an alert up takes the player down too; a dismissal ordered from Dart no longer reports a user exit.
- Seven new localization keys in `app_en.arb`.

## Platform
- [ ] Android
- [x] Android TV
- [ ] iOS
- [x] tvOS
- [ ] Web
[...]

PR #1533, opened 2026-09-13
Fix(windows/android): livetv black picture false unavailable

KevOutIngs:
# Pull Request

## Summary
The tuner status from #1512 called working channels unavailable on Android devices whose decoder renders through a hardware overlay: `PixelCopy` returns black there whatever is on screen, so a healthy channel got the "Channel unavailable" card after 30 s with no way to dismiss it (Fire Cube, Fire Stick).

The picture is now judged from the stream instead of the screen. A black or frozen placeholder compresses to tens of kilobits a second at 1080p where a real channel carries megabits, so the encoded video bits per pixel per second decide, on Android from the frames handed to the decoder and on mpv from the demuxer's `video-bitrate`. A screen read with something in it still proves a picture; a black read proves nothing.

Two more causes of a refused channel change are fixed on every platform: the stop report and live-stream close raced the next PlaybackInfo (a tuner at its stream limit refused the open, Jellyfin 500 in under 300 ms), and the auto-bitrate download could hold a channel change for 8 s on a slow link.

## Related Issues

- Related to #1512 (WizardOfYendor1's report on the Fire Cube)

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

- `PictureEvidence.kt` + `VideoFrameMeter.kt` (Android): a `MediaCodecAdapter` wrapper on the video renderer reports each encoded frame's size and PTS; bits per pixel per second over 1.5 s of media time under 0.1 means nothing to show. Black `PixelCopy` reads are ignored; a non-black read is proof for 10 s. Wire key `pictureBlack` becomes `noPicture`, with a `pictureEvidence` line for the log when the verdict changes.
- `StreamPictureEvidence` (playback_core) + MediaKit backend: the same rule on mpv's `video-bitrate` and frame size, sampledource is open (Windows, Linux).
The tuner status from #1512 called working channels unavailable on Android devices whose decoder renders through a hardware overlay: `PixelCopy` returns black there whatever is on screen, so a healthy channel got the "Channel unavailable" card after 30 s with no way to dismiss it (Fire Cube, Fire Stick).

The picture is now judged from the stream instead of the screen. A black or frozen placeholder compresses to tens of kilobits a second at 1080p where a real channel carries megabits, so the encoded video bits per pixel per second decide, on Android from the frames handed to the decoder and on mpv from the demuxer's `video-bitrate`. A screen read with something in it still proves a picture; a black read proves nothing.

Two more causes of a refused channel change are fixed on every platform: the stop report and live-stream close raced the nexs stream limit refused the open, Jellyfin 500 in under 300 ms), and the auto-bitrate download could hold a channel change for
sues

- Related to #1512 (WizardOfYendor1's report on the Fire Cube)

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] UI/UX update
- [ ] Documentation update
- [ ] Build/CI change
- [ ] Other (describe):

## Changes Made

- `PictureEvidence.kt` + `VideoFrameMeter.kt` (Android): a `MediaCodecAdapter` wrapper on the video renderer reports each encoded frame's size and PTS; bits per pixel per second over 1.5 s of media time under 0.1 means nothing to show. Black `PixelCopy` reads are ignored; a
non-black read is proof for 10 s. Wire key `pictureBlack` becomes `noPicture`, with a `pictureEvidence` line for the log whe
- `StreamPictureEvidence` (playback_core) + MediaKit backend: the same rule on mpv's `video-bitrate` and frame size, sampled once a second while a live source is open (Windows, Linux).
- `PlaybackManager`: one `_releaseInBackground` owns every background stop report and live-stream close and remembers the release; a bringup waits for a pending release (4 s cap) before resolving. A live bringup waits at most 1.5 s for the auto-bitrate measurement; a fresh measurement finishes in the background. A refused live channel logs the server's own words.
- `PlaybackApi.bitrateTest` (Jellyfin, Emby): the bitrate probe goes through the server client instead of a bespoke Dio, whiellyfin returned to the bare `X-Emby-Token`.

## Platform
- [x] Android
- [x] Android TV
- [ ] iOS
- [ ] tvOS
- [ ] Web
- [ ] macOS
- [x] Windows
- [x] Linux
- [x] All / Shared code

## Testing

Fire TV Stick 4K Max (Fire OS, Android 9, MediaTek OMX decoders) against a Jellyfin server with a Dispatcharr tuner behind andows 11 with the MediaKit backend. Unit tests: 10 new Kotlin tests (`PictureEvidenceTest`), 9 new Dart tests(`stream_picture_evidence_test`), all 529 playback tests and the Jellyfin/Emby package tests green.

### Test Steps
1. Open a dead channel whose tuner serves its black failover placeholder (audio plays, picture black). Expected: `picture none` in the log 1 to 3 s after the first frame, "the tuner is still trying" at 8 s, "Channel unavailable" with Retry at 30 s. Seen on Android (0.05 bits/px/s) and Windows.
2. Open working channels (1080p25 H.264, 1080p30 HEVC). Expected: playing within 3 s, no spinner, no card over 45 s. Seen on both.
3. Change channel or press Retry right after a stop on a tuner with a stream limit. Expected: the next PlaybackInfo is sent only after the previous stream's release lands; no 500 from the tuner.
4. First tune after launch on a slow link. Expected: PlaybackInfo sent within about 1.5 s even while the bitrate measurement is still downloading.

## Checklist


KevOutIngs:
@WizardOfYendor1 can you confirm it's working also on your device ?

WizardOfYendor1:
> @WizardOfYendor1 can you confirm it's working also on your device ?

I cherry picked them and deployed them. So far seems like its working on my cube. 

I didn't look at the changes but a suggestion.... if you get to the end state dialog, leave an option to dismiss it versus retrying or killing it (back).

KevOutIngs:
> > @WizardOfYendor1 can you confirm it's working also on your device ?
> 
> I cherry picked them and deployed them. So far seems like its working on my cube.
> 
> I didn't look at the changes but a suggestion.... if you get to the end state dialog, leave an option to dismiss it versus retrying or killing it (back).

Ah yes you're totally right, I'm going to include that too in case it happens to other device.

WizardOfYendor1:
More feedback... I wouldn't show text at all when the channel is tuning on its 1st attempt. Spinner is fine. Show the still trying text after it hits the 1st timeout. i.e. keep the second message, drop the first..... Just my opinion.....

KevOutIngs:
> More feedback... I wouldn't show text at all when the channel is tuning on its 1st attempt. Spinner is fine. Show the still trying text after it hits the 1st timeout. i.e. keep the second message, drop the first..... Just my opinion.....

Alright, sounds good to me, I'm going to do it at the same time. I'm going to ping you again once it's ready.

KevOutIngs:
@WizardOfYendor1 I did the improvement you advice

WizardOfYendor1:
@KevOutIngs  did you mean for this to be ready to merge? Just concerned with other PR not reverted that the next release is going break fire tv users.

Also, have another suggestion (not immediately critical).
The thing I've spent the most time on over the past 1 year+, BY FAR, is live tv tuning speed. You really wouldn't believe the depths I've gone to on this topic.... thus... looking at your changes, I believe what you are doing is measuring/calcuating the bit rate "speed". That value seems to get cached and reused but on a new/cold tune, you wait/block to sample 2.5MB or hit a 1.5 second timeout (???).  On a GBit LAN, meh, it's probably a few tenths of a second but on slow connections it's going to add time to tuning. IF I'm understanding this correctly, I would suggest letting that run as a background task and not block.... get the stream playing ASAP.... then monitor and "help" or report to the user. 


There's also a flag that caught my eye also "vetoesDirectPlay". I after looking, I don't think that's what it's really doing. If not, might be a better, less eye-brow raising name :-).

FYI, Direct Play and Direct Steam  are very confusing names/purposes in jellyfin ( to me anyways).  1st thing is, you MUST have no limits set on your streaming limit in the jellyfin config for you to even have the possibility of getting direct play via live tv. If you have any limit set, you're getting remux'd at a minimum. That means ffmpeg and ffprobe get into play and that means more seconds added to tuning. I have PR in the jellyfin server (unmerged) that goes into the gorey details.  

Also note streams that are MPEG2 video carry bit rate information in their header (mandatory) but that's not true of H.264. Not really relevant here but wanted to mention because of that unavailable info in h.264(5) server side, jellyfin ffprobes it to calculate a bit rate.... sometimes that is wildly inaccurate and thus it gets determined that it exceeds that max and it triggers transcoding. I have comments to that effect in a jellyfin issue also..... but at any rate, I digress.....  :-)

PR #1580, opened 2026-09-17
fix: route game libraries before the collection-type switch

JorisM:
Moonbase lets an admin mark any library as a game library, but the client switches on collectionType first and only reaches the game check in the default branch. So a games library typed books, audiobooks, music or livetv opens the book reader, the music browser or the TV guide instead. On the home screen the same tile already draws a gamepad icon, because the icon does ask isGameLibrary.

This moves the game check in front of that switch, into one libraryRoute() the four entry points share.

Verified against a live Jellyfin 12.1 with a books-typed games library, patched web bundle, nothing else changed: Games opens the games browser => PSP => the game. Before it was /books/<id>?collectionType=books and an empty shelf.

flutter test test/util: 477 passed, 1 failed. That one is game_artwork_scope_sweep_test.dart, which fails the same way on an unmodified parent commit. flutter analyze on the touched files is clean. Reverting only the ordering fails the new test with the live URL:

```
Expected: a string starting with '/games/lib-games'
  Actual: '/books/lib-games?collectionType=books'
```


JorisM:
Closing for now — I want to verify this against a live server before asking anyone to review it. Will reopen once that is done.

PR #1590, opened 2026-09-19
Narrow a long filter list by typing at it

### LINKED ISSUE #1234: [Feature] Ability to search by tags

When you've many tags it gets hard searching through them.

### Proposed solution

 A separate searchbar for tags that autocompletes, for example you writte: "adve" it shows adventure and adventurer, you can select them and they get checked.

(Also right now when selecting multiple tags they get added as an "OR", adding them as "AND" would be cool. I can open another feature for that last one if needed.)

### Platform

All / Not platform-specific

### Alternatives considered

The top right box could be used to search for tags

### Additional context

Image

Kisnov:
# Pull Request

## Summary

A library with hundreds of tags listed every one of them in the filter sheet, so reaching *adventure* meant scrolling past everything alphabetically ahead of it. Any facet longer than 15 values now carries a box above its rows that narrows them as you type — `adve` leaves *adventure* and *adventurer*, and the checkboxes behave as before.

TV side is Moonfin-Client/Smart-TV#439.

## Related Issues

- Closes #1234
- Related to #1589

## Type of Change
- [x] New feature
- [x] UI/UX update

## Changes Made

- `_facetSection` gains a search box for any facet with more than 15 values. Fifteen is roughly where reading the list stops being the quicker option, so short facets are untouched and the sheet looks the same for them.
- Applies to **every** facet, not just Tags. They all go through one `_facetSection`, so restricting it to tags would have meant adding an arbitrary `if (key == 'tags')` to a function that otherwise treats facets alike. Genres on a large library has the same problem. (Confirmed as the preferred scope with the issue author before building.)
- Matching reads the row's **label**, not its value — a language row shows *Catalan* while it filters on `cat`.
- Matching **folds accents both ways**, so a tag answers whether or not the viewer typed its accents: `canco` finds `Cançó`. This uses `foldForSearch` from #1589.
- When a query matches nothing the section says so rather than going silently blank, reusing the existing `noLabelFound` string.
- The logic lives in `lib/ui/util/facet_search.dart` rather than inside the sheet. `_FilterSortDialog` is private and reaching it from a test means standing up the whole navigation chrome (the existing browse widget test needs ~175 lines of fakes for that), while the part worth pinning down is a pure function.
- Adds one string, `searchFacetValues` (`"Search {facet}"`), to `app_en.arb`. The generated `app_localizations_*.dart` files are regenerated and carry the English text until translated, which is why the diff touches 59 files for a one-line string.

## Platform
- [x] All / Shared code

## Testing

- [x] Not tested (explain why): **no manual UI run.** The desktop build needs Visual Studio build tools that are not installed here, so the sheet has not been opened by hand. The logic underneath is unit-tested and the whole suite passes.

Automated, run locally:

- `test/ui/util/facet_search_test.dart` (new) — 9 cases: the threshold either side, the issue's own `adve` example, matching mid-row, blank queries returning the list identically, case-insensitivity, accent folding both directions, and label-vs-value matching.
- Full suite: **2936 passed**.
- `dart analyze`: **0 errors**.

`main` has been merged in to clear a conflict (#1589 landed meanwhile, and its `accent_folding.dart` collided add/add with the copy this branch carried — resolved to upstream's, which adds a `library;` directive). Re-verified after the merge: analyzer still 0 errors.

**One pre-existing failure, not from this PR:** after merging `main`, `test/util/image_cache_index_test.dart` — *"a directory that cannot be created leaves a working index"* — fails on Windows. It arrived in f20d59ee and fails identically on a clean `origin/main` worktree with none of these changes; verified. It expects a `FileSystemException` from creating a directory beneath a regular file, which Windows reports differently. Flagging it rather than quietly ignoring it; happy to open a separate issue.

### Test Steps
1. Open a library with more than 15 tags and open the filter sheet.
2. Expand **Tags** — a search box should sit above the rows.
3. Type `adve` — only *adventure* and *adventurer* remain; tick one and confirm the filter applies.
4. Clear with the ✕ — the full list comes back.
5. Type something no tag matches — the section reads "No Tags found" rather than going blank.
6. Check a short facet (under 15 values) has **no** box.
7. With an accented tag present, confirm the unaccented spelling finds it.
8. Expand **Audio language** and type a language *name* — it should match even though the facet filters on a code.

image
image
## Checklist
- [x] Code builds successfully — `dart analyze` clean; CI builds all platforms on this PR
- [x] No new warnings introduced — 0 analyzer errors

## Not included

The **AND vs OR** behaviour for multiple selected tags, which #1234 raises in passing. That changes how filters are sent to the server rather than the UI, and needs checking against what the Jellyfin and Emby APIs support for tag filtering, so it reviews better separately. The issue author is opening a second issue for it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mattsigal:
Please have Claude adhere to and use the Moonfin PR Memo template and include all requested details (e.g., including screenshots). If you open a new issue in a browser, you can see the template (that your PR memos overwrite). Also, note each repo has it's own variant (so Smart-TV also has a template that should be followed that differs from Moonfin-Core).

Kisnov:
Okey, will do sorry

Kisnov:
## Update: the box could not have been typed into on a set

Device testing of the TV client's equivalent (Moonfin-Client/Smart-TV#439) prompted a check of whether this side had the same problem. It did, from a different cause.

The box went in as a plain `TextField`. This repo ships a fork of a TV text field for exactly this reason, and `LocalSearchField` states it directly:

```dart
/// TV needs [CustomTVTextField] so the remote can reach it and the on screen
/// keyboard opens on select. Everywhere else a plain field with a clear button
/// does the job.
```

It branches on `PlatformDetection.isTV` and, on a set, wraps a `Focus` that opens the keyboard on the select key. A bare `TextField` has none of that, so on **Android TV and tvOS** the box would have been unreachable with the remote and select would not have raised the keyboard.

### Fixed by reusing what is already there

- Dropped the private `_FacetSearchField` and used **`LocalSearchField`**, the field the library header already puts up, so the facet box follows the same platform branch and inherits whatever that field learns later.
- It hardcoded `l10n.searchThisLibrary`, so it now takes an optional `hint` — one line, backwards compatible.
- Each facet keeps a `FocusNode` beside its `TextEditingController`, both disposed.

Net **-46 lines**, and one less place that has to remember TV exists.

### Checks

`dart analyze` **0 errors**. Full suite **2953 passed, 1 failed** — the failure is `image_cache_index_test.dart`, the pre-existing Windows-only one filed as #1591, unrelated to this PR.

### Still unverified

I cannot run this on a set. It is now the same widget the header search uses on Android TV, which is good evidence, but the keyboard actually opening inside the filter sheet has not been seen working. Worth a look from anyone who can sideload an Android TV or tvOS build before merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

